PR #9, opened 2017-01-10
DFTB+ interface and documentation added

patrickmelix:
Added a DFTB+ interface and its documentation. Not all functionality of DFTB+ can be reproduced yet, but used it for several hundred molecular calculations. Hope this is OK for a commit, it's my first one so please tell me if I did something wrong.


patrickmelix:
Sorry, something went wrong moving my changes from master to devel... Will push the changes when it works again. Sorry for the spam.

PR #31, opened 2018-01-30
Crystal Interface

patrickmelix:
This is an interface for the CRYSTAL program. It uses ASE to support molecule-parsing, but also brings a function to do it if there is no ASE installed.
Documentation included.


patrickmelix:
Btw: The CRYSTAL interface in ASE is just some weeks old and not included in the latest release yet. So in order to use it get the latest trunk.

h4nsu:
The change in __init__.py is completely not needed, but besides that: thanks, great job!

patrickmelix:
Not needed, but nice ;-) Thanks!

PR #32, opened 2018-02-02
Add more to GridRunner definitions in plams_defaults

patrickmelix:
The definitions in plams_defaults so far are not enough to reliably run on PBS and SLURM. Added the standard options for allocation of nodes, CPUs and memory so users can acces clusters without changing anything.
Information is taken directly from the man pages of srun and qsub.


h4nsu:
Thanks for another nice contribution, but this time I won't merge it in. `plams_defaults` is supposed to be the place for every user's personal adjustments and preferences. The changes you propose are these kind of adjustments. I'd like to keep the GridRunner definitions there rather minimal and symmetric, meaning that the `special` section contains only the most essential keywords and the same ones for each type of runner.

You are encouraged to use and expand your own `plams_defaults`. If it bothers you that changes done there are constantly visible in `git diff` and `git status`, move your copy of `plams_defaults` somewhere else and point `$PLAMSDEFAULTS` to it.

patrickmelix:
Mmmh, I can understand your goal of keeping it clean. But I do not agree with the conclusion. Without the options the PBS runner is not usable on most systems, since almost no job will use the "standard" allocations of tasks, CPUs and memory. But OK. I'll use the $PYTHONPATH for myself then.

h4nsu:
That's a good point. What if we added a `memory` key to special:

`config.gridrunner.pbs.special.memory = '-l mem='
`config.gridrunner.slurm.special.memory    = '--mem='

patrickmelix:
That would definitively be a good thing, I would agree to that as a comprimise ;-)

h4nsu:
Already in master: e81d454fc3a9c41ef82a9c6d885118b74372bf70 :) Thanks again!

PR #36, opened 2018-03-12
Read/Write Molecules using ASE engine

patrickmelix:
Hey,
this adds an option to use the ASE.io module instead of the PLAMS readers. I think that this is super useful as the ASE projects supports loads of formats that PLAMS will never support and also doesn't have to. In principle this was already possible with the functions from tools.ase, but I think this is a much more straight-forward way to add support for loads of formats. Plan is to have the same for the write function later.

Also the read_molecules function now passes on options to the Molecule.__init__ to enable this new feature.

Added the stuff to the doc so that people can read about it too.

Any suggestions for changes are welcome as always.

Cheers,
Patrick


patrickmelix:
Added writing support and its documentation too.

patrickmelix:
Updated the crystal and dftb+ interfaces to use the new functionality.

h4nsu:
Hi Patrick,
It is indeed a super useful feature, but I don't like the way it's implemented. I would like the core part of PLAMS to remain as independent as possible, with features relying on external libraries being optional plugins. In other words: no ASE (or RDKit, or whatever) in `basemol.py` or anywhere else in `plams.core`. Do you have an idea how this could be implemented in, for example, `ase.py`?

patrickmelix:
Hey,

would be nicer, I agree. At the moment I cannot think of a way to not touch the `basemol.py` while keeping the possibility to use the same syntax for ASE and non-ASE read/write.

If we overwrite the `Molecule.__init__`, `Molecule.read` and `Molecule.write` functions completely when the `ase.py` is loaded one would also loose the PLAMS functionality.... Is there anything these can do that ASE cannot? But this would still not be nice, since the user would never be sure what they are using at the moment and which functionality is available.

Otherwise one would have to live with a second syntax, but I think in general it would be nicer to have an extensible framework for this kind of thing. Then other codes could also be hooked in easily. Perhaps some option like `driver='bla'`, which then automatically tries to load the appropriate module from `tools.bla`. And if its not there just run the PLAMS code...

I'll think about it, perhaps I'll have an idea.

h4nsu:
Why keeping the same syntax is so crucial for you? One could argue that it's better to have some different method while reading/writing using ASE/RDKit to make sure that the user is aware what is actually happening under the hood.
```
mol = fromASE(aseIO.read(filename))
```
We could even "compress" it to a single function, something like `readASEmol()`

That being said, you don't need to overwrite `read` and `write` methods, you can just decorate them on the fly when ASE is detected:
```
try:
    import ase
    Molecule.read = some_read_decorator(Molecule.read)
    Molecule.write = some_write_decorator(Molecule.write)
except ImportError:
    pass
```

Yet another idea would be to abuse a bit the `inputformat` argument in `Molecule.__init__()`:
```
mol = Molecule(filename, inputformat='ase')
```
and write a separate reader/writer for that. In that case `ase.py` could contain something like:
```
if ase_present:
@add_to_class(Molecule)
def readase(...):
    ...
def writease(...):
    ...
Molecule._readformat['ase'] = Molecule.readase
Molecule._writeformat['ase'] = Molecule.writease
```
I actually like a lot this last idea: it's almost completely transparent for the user, while still leaving all the ASE-related code in one place

patrickmelix:
I like the last one too, nice and easy to use. I'll put it on my ToDo list :wink:

patrickmelix:
Closed because I shouldn't do this anymore or because I should reopen a new MR when I finally do it?

h4nsu:
Closed as in "not active at the moment, moved to some future todo list". When you feel like working on that, just reopen this MR.

patrickmelix:
:ok_hand:

PR #37, opened 2018-04-19
BUG: KFReader crashes if German umlaut in kf-file

patrickmelix:
Sadly, our license prints an Umlaut "ä" (0xe4) to the kf files. On some systems this causes a crash of the KFReader like this:

>   File "/home/patrickm/git/myPlamsFork/src/scm/plams/tools/kftools.py", line 73, in read
    ret = self._get_data(self._read_block(f,i))[vtype-1][vstart-1:]
  File "/home/patrickm/git/myPlamsFork/src/scm/plams/tools/kftools.py", line 150, in _get_data
    contents = self._parse(datablock[hlen:], zip((i,d,s,b),(self.word,'d','s',self.word)))
  File "/home/patrickm/git/myPlamsFork/src/scm/plams/tools/kftools.py", line 139, in _parse
    new = tuple(map(lambda x: x.decode() if isinstance(x,bytes) else x, new))
  File "/home/patrickm/git/myPlamsFork/src/scm/plams/tools/kftools.py", line 139, in <lambda>
    new = tuple(map(lambda x: x.decode() if isinstance(x,bytes) else x, new))
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe4 in position 1249: invalid continuation byte

I managed to fix it using the submitted hack. There is probably a nicer way to do it, but I don't know it. So feel free to fix it differently and think of this as a bug-report.

### DIFF
--- a/src/scm/plams/tools/kftools.py
+++ b/src/scm/plams/tools/kftools.py
@@ -135,7 +135,10 @@ class KFReader(object):
         if step > 0:
             while step <= len(block):
                 new = struct.unpack(str(formatstring), block[:step])
-                new = tuple(map(lambda x: x.decode() if isinstance(x,bytes) else x, new))
+                try:
+                    new = tuple(map(lambda x: x.decode() if isinstance(x,bytes) else x, new))
+                except:
+                    new = tuple(map(lambda x: x.decode("Latin-1") if isinstance(x,bytes) else x, new))
                 ret.append(new)
                 block = block[step:]
         return ret


rlKoekie:
Hi Patrick,
can you check if the following works for you, instead of the try: construct?

new = tuple(map(lambda x: x.decode(errors='ignore') if isinstance(x,bytes) else x, new))

patrickmelix:
Hello Hans,

by including `errors='ignore'` the crash is avoided. But for a reason I seem to be unable to understand, the command `read('General','termination status')` returns `None` instead of "NORMAL TERMINATION", which it does with my version...
The string in `new` is the same, except for the "ä" which is missing.
Here I print both tuples for comparison:
Your code:
> (1, 1, 1, 1, 0, 0, 0, 1, 16.0313, 0.0, 0.0, 8.0, 1.88972612499, 0.017453292519943295, '                                                                                H t21.H                                                                                                                                                         C t21.C                                                                                                                                                         End                                                                                                                                                             end input                                                                                                                                                       TAPE21ADF 2017  RunTime: Apr10-2018 14:11:42  Nodes: 1  Procs: 4                                                                                                      *** (NO TITLE) ***                                                                                                                                              u17234                                                                                                                                                          Prof. Thomas Heine / Site license Universitt Leipzig / Leipzig / GERMANY                                                                                       GEOMETRY OPTIMIZATION                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           UnknownCONVERGED                                                                                                                                                       NORMAL TERMINATION                                                                                                                                              ', 0, 0, 0, 0)

My code:
> (1, 1, 1, 1, 0, 0, 0, 1, 16.0313, 0.0, 0.0, 8.0, 1.88972612499, 0.017453292519943295, '                                                                                H t21.H                                                                                                                                                         C t21.C                                                                                                                                                         End                                                                                                                                                             end input                                                                                                                                                       TAPE21ADF 2017  RunTime: Apr10-2018 14:11:42  Nodes: 1  Procs: 4                                                                                                      *** (NO TITLE) ***                                                                                                                                              u17234                                                                                                                                                          Prof. Thomas Heine / Site license Universität Leipzig / Leipzig / GERMANY                                                                                       GEOMETRY OPTIMIZATION                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           UnknownCONVERGED                                                                                                                                                       NORMAL TERMINATION                                                                                                                                              ', 0, 0, 0, 0)

I checked that this is the only affected tuple by using my try-except statement, and it is the only one.
Any ideas?

h4nsu:
Whoa, it took me some time to figure this out. So, in case you're curious:
line 79:                if len(ret) >= vlen:
is responsible for returning value in read. This condition relies on the fact that we know the length of the variable beforehand and keep reading until this length is met. However, if we remove one character from any string in the given block (for a reason I really don't want to talk about all strings from one data block are concatenaded to one string when reading), this condition is never met. Changing the line to if len(ret) >= vlen-1 fixes the thing, unless you add another umlaut ;)

Bottom line is: Hans' solution breaks quite some things in mysterious and unpredictable ways. Let's stick with try-catch maybe? Also, for a bit more transparent error handling, let's catch only UnicodeDecodeError

patrickmelix:
Thanks for the info and the merge!
Might be that there is somebody named Krüger in some "Universität" somewhere, so I approve of the try-except  ;-)

h4nsu:
I think the general problem is not that umlauts are not UTF-8, but just the umlaut you have happens to be not UTF-8. I'll talk to license people and try to maybe convince them to fix it on our side (send licenses with proper encoding)

PR #38, opened 2018-04-23
Get_Gradients for ADF interface

patrickmelix:
Hey,

I know, it doesn't look like much. But I am working on a parallel Frequency calculator for PLAMS/ASE. To make this easier I wanted to use predefined Functions. This way I can check if a certain type of job can be used for automatic parallel Frequency calculations by checking if it has `get_gradients()`.

I added the `inputOrder()` function to make translating between internal and input ordering easier, since this is now used in more than one case. I also don't see why this should be internal, but if needed I can make it so.

`force = -1.0*np.array(result.inputOrder(result.get_gradients(eUnit='ev',lUnit='Angstrom')))`

Cheers,
Patrick

P.S.: Sorry for the late-night mess with commits :grin:


h4nsu:
I'm going to move `inputOrder()` to `SCMResults`, since this is where `_int2inp()` is defined so it makes more sense to me. Also, I'll rename it to `to_input_order()`.

PR #39, opened 2018-04-24
Frequencies in parallel

patrickmelix:
Since it is not possible to speed up numerical ADF Frequency calculations by doing the single-points in parallel I sat down and came up with a new class derived from MultiJob to do exactly that.

Since ASE already has the intelligence to do this just not the muscle (submitting jobs etc), I made this to make use of everything the two worlds have to offer. See the provided documentation for an example.

ASE needed a small change too, so for reference: https://gitlab.com/ase/ase/merge_requests/810

It works like this, if there are any hints on how to make it better please share :grinning:


patrickmelix:
This of course also relies on https://github.com/SCM-NV/PLAMS/pull/38

patrickmelix:
ASE interface changed, working on an update and some improvements. Please don't merge yet.

h4nsu:
Hi,
Is there any fundamental reason why you want to do it in conjunction with ASE?
The following code does the same thing (I guess) in pure PLAMS:
```
class HessianADFResults(Results):

    def get_hessian(self,symmetric=False):
        if symmetric:
            return (self.hessian + self.hessian.T)/2
        return self.hessian

    def get_mw_hessian(self):
        masses = [at.mass for at in self.job.molecule]
        n = len(masses)
        weights = np.empty((n,n))
        for i,j in product(range(n), repeat=2):
            weights[i,j] = masses[i]*masses[j]
        full = np.kron(weights, np.ones((3,3)))
        return self.hessian/full

class HessianADFJob(MultiJob):
    _result_type = HessianADFResults
    def __init__(self, molecule, step=0.01, **kwargs):
        MultiJob.__init__(self, children={}, **kwargs)
        self.molecule = molecule
        self.step = step

    def prerun(self):
        for n_at, axis, step in product(range(1,len(self.molecule)+1), range(3), [-1,1]):
            vec = [0,0,0]
            vec[axis] = self.step * step
            mol = self.molecule.copy()
            mol[n_at].translate(vec)
            self.children[(n_at, axis, step)] = ADFJob(name=self.name+'.{}.{}.{}'.format(n_at, axis, step), molecule=mol, settings=self.settings.child)

    def postrun(self):
        n = len(self.molecule)
        hess = np.empty((3*n,3*n))
        for n_at, axis in product(range(1,n+1), range(3)):
            v1 = np.array(self.children[(n_at, axis, -1)].results.readkf('GeoOpt', 'Gradients_InputOrder'))
            v2 = np.array(self.children[(n_at, axis, 1)].results.readkf('GeoOpt', 'Gradients_InputOrder'))
            hess[3*(n_at-1) + axis] = Units.convert(v2 - v1, 'bohr', 'angstrom')/(2 * self.step)
        self.results.hessian = hess
```

patrickmelix:
Only reason: I am using ASE and it provides more features on top like IR intensities, vibration visualisation etc. that I want to use.

Therefore I think it is easier to use ASE since one can easily integrate PLAMS with the tools they already provide. Reading a Hessian with ASE would of course be possible, but not as nice as a direct link.

patrickmelix:
The ASE changes have been merged, so this can now be used with ASE. As soon as you are happy with it you can therefore merge (or not).

PR #40, opened 2018-05-16
Load jobs without dill files

patrickmelix:
There is no easy way to analyze a mixture of PLAMS jobs and individual jobs run by hand or other scripts. Since nobody uses PLAMS for all of their jobs (even though we should ;-) ), I need PLAMS to be able to retrieve results from jobs that have not been run with PLAMS.
Sometimes it is also just that the PLAMS script terminates before the queued job gets executed, so no pickling is done and I cannot access the results through PLAMS (this might be fixed way nicer later).

This is a hacky way to do it, I know. But I think it is useful anyway. If you have better ideas how to achieve the same please do change everything. This works fine until then.


h4nsu:
With our new input handling system we have a way now to parse an input file back to the `Settings`. I'm going to redo this "external_load" thing (together with pre-run pickle you asked for some time ago). Will probably need a little help with testing if that works for 3rd party codes (Crystal, DFTB+). But for now I'm going to close this one.

h4nsu:
How about that: https://github.com/SCM-NV/PLAMS/commit/a50ae028a1da6895a09fb2f6606c0861d496f5b7 ??

patrickmelix:
I like it. Would you want to include a load-all kind of functionality too? Helps with loading MultiJob structures... Could just be added to the MultiJob class and call the SingleJob.load_external ?!

h4nsu:
I have the following problem with load-all functionality: in the end it's not much more than a single loop:
```
for file in some_collection:
      JobType.load_external(file)
```
In your implementation you added some customisation options with glob, re and ignore, probably because these were the most useful for you. Someone else would probably like the setup in which you can load jobs of different types, so she would like to have a load-all method that takes a list of pairs (jobType, path). Someone else might need a function that goes through all folders and loads each folder using `load` whenever `.dill` file is present, or `load_external` if not. And so on and so on...

What I'm trying to say is that in the end `load_all_external` is just a single for loop prefixed with some "convenience logic". Making this convenience logic very basic will cause the method not to be useful for most of users ("this is too rigid for me, I want to be able to tweak this one thing"). Making the convenience logic general enough to meet requirements of many different users means making it complicated, with tons of optional arguments of strange form (like a list of pairs (jobType, path)). So in the end noone's gonna use it anyway, and people will write their own ad-hoc versions of `for file in some_collection`.

patrickmelix:
Ok, I get that. I'll put something in my own library. Thx for including this!

PR #48, opened 2018-12-17
Rdkit

BvB93:
- global_minimum() has been moved from scm.plams.interfaces.molecule.rdkit to scm.plams.recipes. PLAMS Job objects (and arguments) can be now be passed, allowing the global minimum search to be conducted at an arbitrary level of theory instead of RDKit UFF.
- New in_ring() function in scm.plams.basemol: Checks if an atom or bond is part of a ring system.
- Properties are now preserved upon the use of to_rdmol() and/or from_rdmol(). Objects are pickled by to_rdmol() if they cannot be stored in their native object type. from_rdmol() will automatically unpickle aforementioned objects.
- The proximityBonding keyword (similar in function to guess_bonds()) is now accessible by readpdb(). Setting it to False should prevent crashes upon importing large molecules.


doc/source/cookbook/global_minimum.rst
```diff
@@ -0,0 +1,64 @@
...
+    s.input.basis.type = 'DZP'
+    s.input.XC.GGA = 'PBE'
+    s.input.NumericalQuality = 'Good'
+
+    mol_in = Molecule('my_mol.xyz')
+    job_kwarg = {'settings': s, 'name': 'my_first_DFTJob'}
+    mol_out = global_minimum(mol_in, job_type=ADFJob, **job_kwarg)
+
+
+Depending on the |Job| class, it may be necessary to manually bind the |get_energy()| and |get_main_molecule()| functions to the jobs matching |Results| class, these two functions being used for reading energies and geometries, respectively. For |ADFJob| this is unnecessary, but |AMSJob| for example does not have the |get_energy()| function assigned by default.
```
h4nsu: Unfortunately `|get_energy()|` won't work as you expect. The `| ... |` syntax works only because of looong list of shortcuts/aliases defined at the end of `doc/source/conf.py`. If something is not there, you need to use the explicit Sphinx syntax that points to exact class/function/method: ``` :meth:`~scm.plams.interfaces.adfsuide.adf.ADFResults.get_energy` ```. Alternatively, you could just put it as regular monospace font, without any link: ``` ``get_energy()`` ```

recipes/global_minimum.py
```diff
@@ -0,0 +1,167 @@
...
+
+    return mol
+
+
+def find_bond(mol, no_h=True, no_ring=True, bond_orders=[1.0]):
+    """
+    Create a list of bonds. Each entry is a tuple with indices of atoms forming a dihedral.
+    Consider only diherdals with axis being a single bond, so that rotation is possible.
+
+    ''mol'' |Molecule|: The input molecule
```
h4nsu: Could you please use the ``` :param bool removeHs: ``` syntax everywhere in this file? Just to stay consistent with the rest?

PR #49, opened 2018-12-19
Read/Write using ASE

patrickmelix:
Following up on the discussion in #36 I implemented read/write routines for the Molecule class with binding decorators.

To get the documentation in the "right place" I had to use a nasty trick (see  interfaces/molecule/ase.py). If you know a better way to achieve this with sphinx please let me know!

Also I wanted to get sphinx to print the available formats from the ``Molecule._readformats`` variable, so that in one place the user can find every available formats, but I failed to find out how. I cannot get sphinx to print that variable. Perhaps somebody else does...


interfaces/molecule/ase.py
```diff
@@ -1,16 +1,84 @@
...
+        aseMol = toASE(self)
+        aseMol.write(f, **other)
+        return
+
+
+    Molecule._readformat['ase'] = Molecule.readase
+    Molecule._writeformat['ase'] = Molecule.writease
+
+
+    #Hack to get documentation at the right place
```
h4nsu: This is a really ugly hack :/
I think it would be much better to just "manually" put these two functions in the `molinterfaces.rst` file using regular sphinx `.. function::` directive instead of injecting it into `.py` file like that. In the end autodoc's `.. autofunction::` (and `.. automodule::`) directives are just wrappers that extract docstrings from source and produce a regular `.. function::` or `.. class::` entries.

core/basemol.py
```diff
@@ -398,14 +398,17 @@ class Molecule (object):
     """
 
 
-    def __init__(self, filename=None, inputformat=None, geometry=1, **other):
+    def __init__(self, filename=None, inputformat=None, geometry=1, read_options=None, **other):
```
h4nsu: Initializing `self.settings` with `**other` seems like a dead and useless feature to me. I thought that maybe it would be nicer to use constructor's `**other` for gathering `read_options`? In that case `geometry` should be one of them. That would make the whole setup a bit cleaner. Whaddya think?

interfaces/molecule/ase.py
patrickmelix: The question I got there: I thought that since the functions are only really existent if ASE is available, that they might not get included in the docs if the docs are built on a machine without ASE? That would be bad, since there would me multiple different docs then... Or am I seeing this wrong?

core/basemol.py
patrickmelix: I think the same, just thought you put it there for a reason. I just dorpped all the `geometry` options and moved them to `other`. I updated the doc too.

Also, since this was bugging me all the time: I defaulted the `frame` option in the readers to zero. Starting a selection with 1 instead of 0 is very unpythonic and confusing in my opinion....

interfaces/molecule/ase.py
h4nsu: I think we're overthinking modularity here ;) Someone building doc locally (srsly, who does that? :P) getting ASE documentation without ASE installed is much much less of a problem than SCM website not having ASE documentation, because something went wrong with ASE on the docbuilding bot.

h4nsu:
I'll take it from here. Thanks!

patrickmelix:
:+1: Thanks!

PR #56, opened 2019-02-20
Cp2k Interface Updates

patrickmelix:
Hey,

<del>**this is work in progress, so please do not merge yet.**</del>

Wanted to get feedback while working to not mess things up. I would especially be interested in @felipeZ's feedback since he wrote the interface.

List of things I did:
* `_h` for the `Settings` parser, just like in the SCM interfaces
* file copying for additional input/molecule files. Usefull since CP2K allows inclusion of files in input.
* `Cp2kJob` is now added to the namespace (wasn't for me before because no `__all__` was set).
* updated the doc, e.g. `settings.input.global.project = 'test'` would not work because nothing can be named `global` and some more things....
* `Cp2kResults` with some functions I need. Also includes `recreate_settings` to load old jobs.

I <del>want to also write</del> wrote a Result class <del>later on</del>, just started working with it and want to use PLAMS to speed up my MD procedures. 

Suggestions and criticism are most welcome.

Cheers,
Patrick

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -282,6 +282,7 @@ rst_epilog = """
 .. |mol2CrystalConf| replace:: :func:`mol2CrystalConf<scm.plams.interfaces.thirdparty.crystal.mol2CrystalConf>`
 
 .. |Cp2kJob| replace:: :class:`Cp2kJob<scm.plams.interfaces.thirdparty.cp2k.Cp2kJob>`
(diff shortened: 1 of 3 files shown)


patrickmelix:
So, from my side this is it. I am using all the new functionality and it seems to work well (for me at least).

Can be merged, please check if the doc is ok. I get a lot of errors from a lot of places when compiling it but I don't think I added any :wink:

h4nsu:
Thanks for another great contribution!

PR #64, opened 2019-04-30
Addd hashing for non-bonded clusters

leobunga:
Check if self.separate() returns more than one molecule. If so, rehash each name.


h4nsu:
Ugh, `separate` can be a really heavy operation, it's better not to do it by default for every `label` call. Especially considering the fact, that `label(level=(1,3,4))` calls `label` 3 times.
And, on top of that, your implementation calls `separate` twice...

PR #66, opened 2019-05-23
Settings update and other minor changes

BvB93:
interfaces.molecule.rdkit:
* Added `'get_substructure'` to `__all__`.
* @robertrueger Added the ``useRandomCoords=True`` argument to the ``AllChem.EmbedMultipleConfs()`` function in ``scm.plams.interfaces.molecule.rdkit.get_conformations()``. Long story short: This should prevent crashes when dealing with large systems.
* @robertrueger Added ``proximityBonding=False`` to ``readpdb()``. Should prevent crashes when importing large molecules.

~~interfaces.thirdparty.cp2k.Cp2kResults:~~
~~* Added the `unit` argument to `Cp2kResults.get_energy()` and `Cp2kResults.get_dispersion()`.~~

mol.molecule.Molecule:
* Added the `atom_subset` argument to `Molecule.guess_bonds()`; this allows the bond guessing to be confined to a subset of atoms.
* Added the `start` argument to `Molecule.set_atoms_id()`, `start` representing the starting value of the enumeration procedure.
* Supplied the `start` argument to a number of [enumerations](https://docs.python.org/3.7/library/functions.html?highlight=enumerate#enumerate) in the `Molecule()` class, thus making a lot of `i+1` statements redundant.
* Added the `Molecule.__array__()` method. This method ensures that passing a `Molecule()` instance to `numpy.array()` produces an array of Cartesian coordinates.

core.settings.Settings:
* The `dict.__getattr__()` method, referenced in `Settings.__getattr__()`, does not exist. It has been replaced with `dict.__getattribute__()`.
* The `sorted` function, called in `Settings.__iter__()`, will raise a `TypeError` if the keys  consist of both strings and integers. This has been fixed by adding the `key=str` to `sorted()`, thus converting all elements into strings during the sorting process.
* While the `Settings.__init__()` method converts `dict[list[dict]]` into `Settings[list[Settings]]`, the opposite does not hold for `Settings.as_dict()`. This has now been fixed.
* Fixed a bug where ``Setings.__setattr__()`` and ``Settings.__delattr__()`` would, respectively, call ``Setings.__setitem__()`` and ``Settings.__delitem__()`` even when magic method are provided.
* Indentation in strings produced by `Settings._str()` is now more consistent.
Should look better overall when processing values containing the newline (``"\n"``) character.
* Added the ``Settings.get_nested()`` and ``Settings.set_nested()`` methods (see Example 1). The two methods can be used to get or set values with:
  * Variable levels of nesting.
  * Variable keys.
  * @mirkofranchini: The default behaviour of ``Settings.__missing__()`` can be supressed during the getting and setting with the `ignore_missing` argument, thus raising a `KeyError` if a key is missing. This should help out with Roberts' problem. See also the new `Settings.enable_missing()` context manager (see Example 3).  
* Added the `Settings.flatten()` and `Settings.unflatten()` methods (see Example 2). The two methods can be used for interconverting between flattened and nested dictionaries, respectively.

Examples:
* Example 1:

         >>> s = Settings()
         >>> print(s)

         >>> s.set_nested(('a', 'b', 'c'), True)
         >>> print(s)
         a:      
           b:    
             c:  True

         >>> value = s.get_nested(('a', 'b', 'c'))
         >>> print(value)
         True  

* Example 2:

         >>> s = Settings()
         >>> s.a.b.c = True
         >>> print(s)
         a:      
           b:    
             c:  True

         >>> s_flat = s.flatten()
         >>> print(s_flat)
         ('a', 'b', 'c'):        True

         >>> s_unflat = s_flat.unflatten()
         >>> print(s_unflat)
         a:      
           b:    
             c:  True

* Example 3:

         >>> s = Settings()
         >>> with s.enable_missing():  # Attempt #2 with Settings.__missing__ disabled
         >>>     s.a.b.c = True
         KeyError: 'a'     

         >>> s.a.b.c = True  # Attempt #2 with Settings.__missing__ enabled
         >>> print(s.a.b.c)
         True


BvB93:
Reverted the additions to interfaces.thirdparty.cp2k.Cp2kResults, see https://github.com/SCM-NV/PLAMS/pull/67.

BvB93:
> @mirkofranchini: The default behaviour of ``Settings.__missing__()`` can be supressed during the getting and setting with the ignore_missing argument, thus raising a KeyError if a key is missing. This should help out with Roberts' problem.

Another, more general, solution would be to implement a context manager for altering the behaviour of ``Settings.__missing__()`` (see example 3). Any thoughts on this subject?

Example 3:

         >>> s = Settings()
         >>> s.a.b.c = True

         >>> with s.EnableMissing(s):  # Temporary revert from Settings.__missing__ to dict.__missing__
         >>>     s.d.e = True
         KeyError: 'd'     

         >>> s.d.e = True  # Attempt number two with Settings.__missing__
         >>> print(s.d.e)
         True

BvB93:
@robertrueger 

interfaces.molecule.rdkit:
* Added the ``useRandomCoords=True`` argument to the ``AllChem.EmbedMultipleConfs()`` function in ``scm.plams.interfaces.molecule.rdkit.get_conformations()``. Long story short: This should prevent crashes when dealing with large systems.

core/settings.py
```diff
@@ -216,20 +220,193 @@ def as_dict(self):
...
+            >>> value = s.get_nested(('a', 'b', 'c'))
+            >>> print(value)
+            True
+        """
+        s = self
+        if ignore_missing:
+            for k in key_tuple:
+                s = s[k]
+        else:  # Ignore Settings.__missing__ and raise a KeyError if a key is missing
+            with s.enable_missing():
```
h4nsu: To avoid code duplication here and in `set_nested` it would be better to do something like:
```
with contextlib.supress() if ignore_missing else s.enable_missing():
            for k in key_tuple:
                    s = s[k]
```
See more: https://stackoverflow.com/questions/45187286/how-do-i-write-a-null-no-op-contextmanager-in-python

core/settings.py
```diff
@@ -216,20 +220,193 @@ def as_dict(self):
...
+        """
+        return EnableMissing(cls)
+
+
+    def get_nested(self, key_tuple, ignore_missing=True):
+        """Retrieve a nested value by, recursively, iterating through this instance using the keys in *key_tuple*.
+
+        The :meth:`.Settings.__getitem__` method is called recursively on this instance until all keys in key_tuple are exhausted.
+
+        Setting *ignore_missing* to ``False`` will internally open the :meth:`.Settings.enable_missing` context manager, thus raising a :exc:`KeyError` if a key in *key_tuple* is absent from tihs instance.
```
h4nsu: tihs -> this

core/settings.py
```diff
@@ -216,20 +220,193 @@ def as_dict(self):
...
+            elif isinstance(v, list):
+                d[k] = [i.as_dict() if isinstance(i, Settings) else i for i in v]
+            else:
+                d[k] = v
 
         return d
 
 
+    @classmethod
+    def enable_missing(cls):
```
h4nsu: This thing is super cool, but please please change the name ;P If you call it "a context manager for temporary disabling the `__missing__` magic method", then why on earth is it called `enable_missing`? ^^

`disable_autocomplete` maybe? or just `disable_missing`?

BvB93:
The requested changes have been addressed in https://github.com/SCM-NV/PLAMS/pull/66/commits/63450be7da4516580c41c8eb6cd7ab4976a9d61d.

PR #67, opened 2019-05-24
CP2K enhancements

patrickmelix:
Hey,

just saw that there are some minor changes in #66 that affect the CP2K interface I'm also still working on. Sorry for not putting that here earlier, didn't know somebody else is working on it too and was busy with something else in the meantime.

I also added the unit option for the energies here, so this might conflict with #66. Also there's now some things around MDs.

~~Not final yet, feel free to request changes (especially @BvB93 ).~~
Cheers,
Patrick


BvB93:
Hi Patrick,

The changes you've made look just fine to me.
As you've already added the `unit` keyword in your branch, i'll remove it from https://github.com/SCM-NV/PLAMS/pull/66 for the sake of minimizing redundancy.

Regards, Bas

patrickmelix:
Awesome, thanks. And sorry for not pushing earlier...

patrickmelix:
I think nothing more is coming any time soon from me. Sorry for the delay. Could be merged.

interfaces/thirdparty/cp2k.py
```diff
@@ -140,25 +142,72 @@ def get_runtime(self):
...
-            select = -1
-        s = self.grep_output(search + ' energy:')[select].split()[-1]
-        return float(s)
+            select = slice(0,None)
 
-    def get_energy(self, index=0):
-        """Return last occurence of 'Total energy:' in the output."""
-        return self._get_energy_type('Total', index=index)
+        s = self.grep_output(search+' energy:')[select]
+        if isinstance(select, int):
```
BvB93, line 152:
```suggestion
        if not isinstance(select, slice):
```

I'd recommend swapping things around here.

In its current state the check will fail with, for example, the likes of `numpy.integer` or any other object which defines a `__index__` method.

patrickmelix:
Cleaning my ToDo-List: Anything more for me to do @BvB93?

interfaces/thirdparty/cp2k.py
```diff
@@ -140,25 +142,82 @@ def get_runtime(self):
...
+        chunk = self.get_output_chunk(begin="SUBROUTINE",end="-"*70)
+        ret = {}
+        for line in chunk[1:]:
+            line = line.split()
+            ret[line[0]] = line[1:]
+        return ret
+
+
+    def _get_energy_type(self, search='Total', index=-1, unit='a.u.'):
         if index:
```
BvB93, line 156: WIll this behave as intended when `index=0`?
Considering `bool(0) is False`.

interfaces/thirdparty/cp2k.py
```diff
@@ -140,25 +142,82 @@ def get_runtime(self):
...
+        """Returns 'Total energy:' from the output file.
+
+        Set ``index`` to choose the n-th occurence of the Charges in the output, e.g. to choose an optimization step.
+        Also supports slices.
+        Set to *None* to return all as a list.
+        Defaults to the last occurence.
+
+        Returns a List of Energies.
+        """
+        return self._get_energy_type('Total', index=index, unit=unit)
```
BvB93, line 176: I don't think returning a list of energies is a good idea here, as all the other `Job.get_energy()` methods return scalars.
Is it an idea to an additional keyword argument?

BvB93 (commented):
A few, mostly style-related, comments but overall I feel these are nice additions.
The only thing that I find a bit concerning is `Cp2kResults.get_energy()` returning a list rather than a float, as it is the only `Job.get_energy()` which does so.

Secondly, Michal just said that he'll take a look at the open PRs, which should help with the whole merging thing.

interfaces/thirdparty/cp2k.py
patrickmelix: True, fixed this. No need for this check anyhow.

interfaces/thirdparty/cp2k.py
patrickmelix: True, changed the docstrings and the `_get_energy_type` returns.

patrickmelix:
Thx for the input @BvB93, very good!

interfaces/thirdparty/cp2k.py
```diff
@@ -252,6 +303,107 @@ def get_multigrid_info(self):
...
+            elif len(frame) == 0:
+                continue
+            else:
+                ret.append(frame)
+
+        names = [ x[0] for x in ret[0] ]
+        for i, frame in enumerate(ret):
+            assert names == [ x[0] for x in frame ], ("Namings in output not constant?")
+            assert len(frame) == len(names), ("{:}".format(frame))
+            assert set([len(x) for x in frame]) == set([2])
```
BvB93, line 347:
```suggestion
            assert set(len(x) for x in frame) == {2}
```
Nit, feel free to ignore: 
`set()` can take any iterable object as long as its element are hashable. 
This includes both lists and generators.

BvB93 (commented):
Really liking the changes!
I've got a few more documentation-related questions/remarks but in my opinion the PR is good to go afterwards.

interfaces/thirdparty/cp2k.py
patrickmelix: Ahh, didn't know. Thx!

interfaces/thirdparty/cp2k.py
```diff
@@ -140,25 +142,74 @@ def get_runtime(self):
...
+        for line in block:
+            line = line.strip().split()
+            if len(line) == 0:
+                continue
+            #new forces block
+            if line[0] == '#':
+                ret.append([])
+                continue
+            ret[-1].append(line[-3:])
+        return [np.array(item, dtype=float) for item in ret]
```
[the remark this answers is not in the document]
patrickmelix: I guess in almost all cases they will be the same length. But in any simulation without constant number of particles, the array length might differ between frames. Not sure if there is any calculation in CP2K that allows this (perhaps some embedded schemes?!), but I guess this way we are on the safe side...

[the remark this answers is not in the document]
BvB93: > I guess in almost all cases they will be the same length.

By the sound of it the this would entail the vast majority (all?) cases.

How about converting the whole thing into an `ndarray` via a `try` / `except` approach,
using a ragged array as fallback if necessary?

``` python
import warnings
import numpy as np

def _sequence_to_array(seq, dtype=float):
    """Convert the (nested) sequence *seq* into an array with the specified *dtype*.
    
    If *seq* is ragged sequence it will be converted into a ragged object array instead.
    """
    try:
        return np.array(seq, dtype=dtype)
    except ValueError as ex:
        warning = RuntimeWarning(f"Failed to coerce the forces into a single {np.dtype(dtype)} array;"
                                 "returning a ragged array instead")
        warning.__cause__ = ex
        warnings.warn(warning, stacklevel=3)
        
        # Create and return a ragged array, 
        # i.e. an object array whose elements are normal arrays
        ret = np.empty(len(seq), dtype=object)
        ret[:] = [np.array(item, dtype=dtype) for item in seq]
        return ret

class Cp2kResults:
    get_forces(...):
        ...
        return _sequence_to_array(ret)
```

BvB93:
@felipeZ do you have any comments? 
Considering this entails a `Cp2kJob` / `Cp2kResult` enhancement.

PR #78, opened 2020-01-31
Added the Molecule.index(), Molecule.get_integer_bonds() & Molecule.round()

BvB93:
Added three new methods: ``Molecule.index()``, ``Molecule.get_integer_bonds()`` and ``Molecule.round()``.
* ``Molecule.index()``: Return the index/indices of the passed [``Atom``](https://www.scm.com/doc/plams/components/atombond.html#atom)/[``Bond``](https://www.scm.com/doc/plams/components/atombond.html#bond) akin to [``list.index()``](https://docs.python.org/3/tutorial/datastructures.html#more-on-lists). 
  For example: ``1 == mol.index(mol[1])`` and ``(1, 2) == mol.index(mol[1, 2])``. 
* ``Molecule.set_integer_bonds()``: Convert non-integer bond orders into integers.
  The function walks through a set of graphs containing all non-integer bonds, alternating calls to [``math.ceil()``](https://docs.python.org/3/library/math.html#math.ceil) and [``math.floor()``](https://docs.python.org/3/library/math.html#math.floor).
* ``Molecule.round()``: Round the Cartesian coordinates of molecule to a given precision in decimal digits.
* Generalized [``Molecule.from_array()``](https://www.scm.com/doc/plams/components/mol_api.html#scm.plams.mol.molecule.Molecule.from_array) such that it accepts any iterable consisting of xyz coordinates.


mol/molecule.py
```diff
@@ -604,6 +605,214 @@ def unit_cell_volume(self, unit='angstrom'):
...
+            """Depth-first search algorithm for integer-ifying the bond orders."""
+            for b2 in atom.bonds:
+                if b2._visited:
+                    continue
+
+                b2._visited = True
+                b2.order = func(b2.order)  # func = ``math.ceil()`` or ``math.floor()``
+                del bond_dict[b2]
+
+                atom_new = b2.atom1 if b2.atom1 is not atom else b2.atom2
```
h4nsu: `Bond.other_end` could be used here

unit_tests/test_identify_CO_6.py
```diff
@@ -1,4 +1,5 @@
-from scm.plams import Molecule, label
+from scm.plams import Molecule
+from scm.plams.mol.identify import label
```
h4nsu: I'm not sure if this import is needed, as `label` is a method of `Molecule`. Same for all other tests in `unit_tests`

BvB93: Same testing showed that, unless explictly imported, the ``label()`` method will not be automatically attached to the ``Molecule`` class. Maybe ebcause it's not specified in ``__all__``?

mol/molecule.py
BvB93: Good call!

BvB93: See https://github.com/SCM-NV/PLAMS/pull/78/commits/170e1e5bcc4431aaa18a89b5af9a2b3e38ad8fce.

unit_tests/test_identify_CO_6.py
h4nsu: Seems super strange, but ok, if it breaks without it... ;)

core/functions.py
```diff
@@ -313,3 +315,33 @@ def parse_heredoc(bash_input: str, heredoc_delimit: str = 'eor') -> str:
...
+    * ``action = 'ignore'``: Do nothing.
+    * ``action = 'warn'``: Issue a warning.
+    * ``action = 'raise'``: Raise the passed exception.
+
+    """
+    try:
+        return _ACTION_DICT[action]
+    except KeyError as ex:
+        raise ValueError("`action` expected 'ignore', 'raise' or 'warn'; "
+                         f"observed value: {action!r}") from ex
```
h4nsu: Maybe `core/private` is a better place for that?

BvB93: And done https://github.com/SCM-NV/PLAMS/pull/78/commits/67cb434ce55ee887b989e8273c346d49b3a19405.

PR #81, opened 2020-03-09
Fixed an issues where pickling and deep copying Molecules could raise RecursionErrors

BvB93:
* Added the ``Molecule.__getstate__()`` and ``Molecule.__setstate__()`` methods in order to circumvent any [``RecursionError``](https://docs.python.org/3/library/exceptions.html#RecursionError); see [Pickling Class Instances](https://docs.python.org/3/library/pickle.html#pickling-class-instances).
* Backwards compatibility is retained with respect to previously pickled Molecules.
* Bonus points: pickling is now ~25% faster as observed with systems containing both ``5`` and ``17 000`` atoms.
* Added the ``Molecule.__deepcopy__()`` method which, similar to the pickling changes, means [``copy.deepcopy()``](https://docs.python.org/3/library/copy.html#copy.deepcopy) can now be used without raising a [``RecursionError``](https://docs.python.org/3/library/exceptions.html#RecursionError).


BvB93:
Todo: add a few unit tests.

Issue #82, opened 2020-03-12
Allow multiple CP2K executables


felipeZ:
Currently, `cp2k.popt` is used for running CP2K. It would be great if the user can optionally pick another executable [from the available binaries](https://github.com/cp2k/cp2k/blob/master/INSTALL.md#3-compile)

h4nsu:
@felipeZ I assume I can close this issue now, right?

felipeZ:
Sorry, I forgot to close it!

PR #83, opened 2020-03-12
allow other CP2K executables

felipeZ:
solved #82

### DIFF
--- a/interfaces/thirdparty/cp2k.py
+++ b/interfaces/thirdparty/cp2k.py
@@ -1,6 +1,7 @@
 """Class to manipulate CP2K jobs."""
 import subprocess
 import shutil
+from pathlib import Path
 from os.path import join as opj
 
 from ...core.basejob import SingleJob
@@ -368,7 +369,7 @@ class Cp2kJob(SingleJob):
         cp2k_command = self.settings.get("executable", "cp2k.popt")
 
         # Check the executable name
-        available_executables = (
+        available_executables = set({
             # Serial single core testing and debugging
             "sdbg",
             # Serial general single core usage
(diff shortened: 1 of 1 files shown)


interfaces/thirdparty/cp2k.py
```diff
@@ -340,64 +343,83 @@ def parse(key, value, indent=''):
...
+            "sopt",
+            # Parallel (only OpenMP), single node, multi core
+            "ssmp",
+            # Parallel (only MPI) multi-node testing and debugging
+            "pdbg",
+            # Parallel (only MPI) general usage, no threads
+            "popt",
+            # parallel (MPI + OpenMP) general usage, threading might improve scalability and memory usage
+            "psmp")
+        if not any((f"cp2k.{x}" == cp2k_command.lower() for x in available_executables)):
```
BvB93: I'd recommend changing ``available_executables`` to a [``set``](https://docs.python.org/3/library/functions.html#func-set) and then only checking the extension of ``cp2k_command``:

```
available_executables: set = {...}

executable = cp2k_command.rsplit('.')[-1]
if executable not in available_executables:
    raise RuntimeError(...)
```

Membership checking (*i.e.* ``set.__contains__()``) is quite a bit faster with sets as compared to builtin Sequence types.

interfaces/thirdparty/cp2k.py
```diff
@@ -340,64 +343,83 @@ def parse(key, value, indent=''):
...
+            msg = f"unrecognized cp2k executable: {cp2k_command}"
+            raise RuntimeError(msg)
+
         # Try to run cp2k using mpirun and otherwise srun (if available)
         command_tuple = ('mpirun', 'srun')
```
BvB93: It's probably best to use ``command = ('',)`` if one of the single-core options is used; it might cause issues otherwise.

interfaces/thirdparty/cp2k.py
BvB93: It might be better btw to use [``os.path.splitext``](https://docs.python.org/3.8/library/os.path.html#os.path.splitext) instead of [``str.rsplit``](https://docs.python.org/3/library/stdtypes.html#str.rsplit), as the former can handle any path-like object:

```
_, executable = os.path.splitext(cp2k_command)
```

interfaces/thirdparty/cp2k.py
felipeZ: Right, I will change it!

interfaces/thirdparty/cp2k.py
felipeZ: thanks!

interfaces/thirdparty/cp2k.py
```diff
@@ -368,7 +369,7 @@ def get_runscript(self):
         cp2k_command = self.settings.get("executable", "cp2k.popt")
 
         # Check the executable name
-        available_executables = (
+        available_executables = set({
```
BvB93: The ``set({...})`` syntax is redundant btw, ``{...}`` already creates a set.

interfaces/thirdparty/cp2k.py
```diff
@@ -340,64 +344,87 @@ def parse(key, value, indent=''):
...
+            ret = cp2k_command
+        else:
+            for command in ('mpirun', 'srun'):
+                try:
+                    subprocess.run([command, "--help"],
+                                   stdout=subprocess.DEVNULL)
+                    ret = f"{command} {cp2k_command}"
+                    break
+                except OSError:
+                    pass
```
BvB93: It's probably wise to raise a ``RuntimeError`` if the ``break`` statement is never reached:
```
            for command in ('mpirun', 'srun'):
                try:
                    subprocess.run([command, "--help"],
                                   stdout=subprocess.DEVNULL)
                    ret = f"{command} {cp2k_command}"
                    break
                except OSError:
                    pass
            else:
                raise RuntimeError("Neither 'mpirun' nor 'srun' were successfully found")
```

Issue #80, opened 2020-03-05
Settings.ignore_missing() and multi-threading


BvB93:
``Settings.ignore_missing()`` being called by multiple threads can lead to some rather problematic behaviour.

The example below will work just fine:
```
>>> s = Settings(...)
>>> with s.ignore_missing():
...     results = job.run(settings=s, ...)
```

But if ``Settings.ignore_missing()`` is called somewhere down the stack of ``Job.run()`` then  exiting the context manager in thread ``#i`` *will* do the same in all other threads (as ``Settings.ignore_missing()`` modifies the ``Settings`` class, not the instance).

BvB93:
Some tweaking of the [``reprlib.recursive_repr()``](https://github.com/python/cpython/blob/6b452ff97f70eca79ab956987cc04b6586feca00/Lib/reprlib.py#L9) decorator might do the trick.

BvB93:
So it seems like this problem is much easier to solve with the help of [``threading.Lock``](https://docs.python.org/3/library/threading.html#threading.Lock), which does *exactly* what is required here: waiting until all threads are synchronized before continuing.
In any case, I've created a pull request at https://github.com/SCM-NV/PLAMS/pull/84.

PR #84, opened 2020-03-12
Settings update

BvB93:
See https://github.com/SCM-NV/PLAMS/issues/80.

Changes
---------
* Added a threading safeguard to the ``SupressMissing`` context manager using [``threading.Lock``](https://docs.python.org/3/library/threading.html#threading.Lock).
* Added a single ``SupressMissing`` instance to ``Settings`` as a class variable. ``Settings.supress_missing()`` wil therefore now always return a singleton.
* Replaced all explicit references to ``Settings`` with ``cls = type(self)``. 
This should make subclassing much easier by treating the class type as a variable.
* Replaced ``dict.func_name`` with ``super().func_name``.
* Added tests for the thread-safeguard functionality.
* Raise a ``ReentranceError`` when calling ``SupressMissing.__enter__()`` in a [reentrant](https://docs.python.org/3/library/contextlib.html#reentrant-context-managers) manner.
* Created a separate module for context managers (``scm.plams.core.context``).

### DIFF
--- a/core/settings.py
+++ b/core/settings.py
@@ -1,61 +1,17 @@
 import textwrap
-import threading
 import contextlib
-from functools import wraps
 
-from .errors import ReentranceError
+from .context import SupressMissing, Lower, Upper
 
 __all__ = ['Settings', 'ig']
-
-
-class SupressMissing(contextlib.AbstractContextManager):
-    """A reusable, but non-reentrant, context manager for temporary disabling the :meth:`.Settings.__missing__` magic method.
-
-    See :meth:`Settings.supress_missing` for more details.
-
-    """  # noqa
-
-    def __init__(self, obj: type):
-        """Initialize the context manager."""
-        self.lock = threading.Lock()
-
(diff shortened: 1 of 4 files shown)


core/settings.py
```diff
@@ -301,10 +354,11 @@ def set_nested(self, key_tuple, value, supress_missing=False):
                 c: 	True
         """
         s = self
+        *initial_keys, final_key = key_tuple
```
BvB93: Rewrote this such that ``key_tuple`` can be any type of iterable and not just a sequence.

h4nsu:
Honestly, I think this is a bit too much complication for having a fairly exotic and rather uncommon functionality. Maybe it's just better to leave it the way it is now? And jusr put a warning that suppress_missing shouldn't be used in parallel context with caution.

When it comes to case-insensitivity I found a different, much simpler solution. See !85

BvB93:
That's fair, it's a rather niche case anyway as you've noted.

PR #85, opened 2020-04-23
Make Settings Great Again!

h4nsu:
Settings are now case-preserving but case-insensitive


mirkofranchini:
Awesome! (passes all tests using PLAMS in the AMS trunk).

PR #90, opened 2020-06-17
Use srun by default in CP2K

felipeZ:
fix #86

### DIFF
--- a/interfaces/thirdparty/cp2k.py
+++ b/interfaces/thirdparty/cp2k.py
@@ -382,27 +382,23 @@ class Cp2kJob(SingleJob):
             "popt",
             # parallel (MPI + OpenMP) general usage, threading might improve scalability and memory usage
             "psmp"}
-        suffix = Path(cp2k_command).suffix[1:]
-        if suffix not in available_executables:
-            msg = f"unrecognized cp2k executable: {cp2k_command}"
-            raise RuntimeError(msg)
 
-        # Try to run cp2k MPI binaries using mpirun and otherwise srun (if available)
-        if suffix in set({'sdbg', 'sopt'}):
-            ret = cp2k_command
-        else:
-            for command in ('srun', 'mpirun'):
-                try:
-                    subprocess.run([command, "--help"],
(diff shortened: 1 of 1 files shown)


BvB93:
While you're at it, could you add an `else` statement after the `for` / `break` loop?

If the `break` statement is never reached then it will get stuck on `command = "mpirun"`, even though mpirun is unavailable in such case. This is guaranteed to crash the job down the line.

``` python
for command in ('srun', 'mpirun'):
    try:
        ret = ...
    except OSError:
        break
else:  # In case the previous `break` statement is never reached
    ret = 'sopt'
```

felipeZ:
@BvB93 you are write :) I have clean a bit the code and allow the user to invoke her/his own call to MPI using `srun` or `mpirun`

interfaces/thirdparty/cp2k.py
```diff
@@ -382,27 +382,24 @@ def get_runscript(self):
...
+            suffix = Path(cp2k_command).suffix[1:]
+
+            if suffix not in available_executables:
+                msg = f"unrecognized cp2k executable: {cp2k_command}"
+                raise RuntimeError(msg)
+
+            # Try to run cp2k MPI binaries using mpirun and otherwise srun (if available)
+            if suffix not in set({'sdbg', 'sopt'}):
+                available = (shutil.which(c) for c in available_mpi_commands)
+                mpi_command = next((c for c in available if c is not None), "")
```
BvB93, line 400: Even when `mpi_command = ""` then `cp2k_command` still be set to `"cp2k.popt"` by default (which requires `mpirun` or `srun` to properly function I believe?).

felipeZ: Actually `cp2k.popt` does not strictly required the use of mpirun, these two command are equivalent:
```bash
mpirun -np 1 cp2k.popt -i input.inp -o output.out
cp2k.popt -i input.inp -o output.out
```

BvB93: Ah ok, nevermind then.

PR #93, opened 2020-07-29
Raspa Interface

patrickmelix:
Wanted to run some RASPA2 calculations.

The order of third-party interfaces bugged me. Ordered them alphabetically.

Added the Kelvin (K) energy unit, since it is heavily used in RAPSA2

~~Work in Progress: Please do comment on the current status, but don't yet merge.~~


patrickmelix:
Works fine for me, no more work planned for now.

PR #95, opened 2020-10-19
FIX: Loading a Job with Ghost Atoms

patrickmelix:
Loading an ADF job with ghost atoms (e.g. `Gh.C`) is not working, because the periodic table module does not support ghosted atoms. I hacked an `elif` clause. I think it is the best way to go, even though the elements are lost this way (replaced by `Xx`, the standard PLAMS special atom).

Passing the element would end up with a complete molecule, which is I think much more prone to misinterpretation.

### DIFF
--- a/interfaces/adfsuite/adf.py
+++ b/interfaces/adfsuite/adf.py
@@ -192,10 +192,10 @@ class ADFResults(SCMResults):
                     with InputParser() as parser:
                         inp = parser.to_settings('adf', new_input)
                 except:
-                    log('Failed to recreate input settings from {}'.format(self._kf.path, 5))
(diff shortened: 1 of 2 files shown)


patrickmelix:
After discussions with @robertrueger, this looks like it is not an issue in AMS2020 anymore. For compatibility with previous releases I suggest to merge this small fix, so that one can at least load previous ADF results using `load_external`.

patrickmelix:
Bad style to fix more than one issue, but the log messages really annoyed me and it was such a quick fix...

h4nsu:
No worries, nice catch with the logs. Anyway, ADFJob is slowly becoming obsolete, but it's still nice to have such little things fixed. Thanks!

PR #97, opened 2020-11-09
Basejob: load_external, New Jobname Option

patrickmelix:
Loading jobs that have been manually run often requires renaming files to fit the PLAMS conventions. By introducing this small change, at least the folder can be named differently than the files.
This also fits better with the AMS-GUI behaviour, files are automatically named as in PLAMS but the folder name is chosen manually.

I know that this can break some things, since the job is then not in the proper place. But I think if someone finds this option they know what they are doing...


h4nsu:
Meh, I have mixed feelings about that :/ It seems like it explicitly gives user a possibility to break one of the main invariants of the whole environment. I need to think about it...

BTW it's quite likely that there are going to be some changes in the job management in PLAMS soon (to make it a bit more interactive), so maybe that fix won't be needed in the end.

patrickmelix:
I already expected that since we once had a discussion about something similar... My argument would be, that load_external does not happen in the working folder, so there will be no conflict with newly written files etc. It allows reading jobs outside the work environment and create jobs based on that following the standard PLAMS logic. So I could not think of a way to use this change to break anything, even intentionally. But perhaps you can. If so, let me know and we might find a way to prevent this.

I really think it is necessary to have this feature, not only for my way of working. In my experience everybody starts with manual calculation before using PLAMS. So interaction of PLAMS with manual jobs is kinda crucial.

Is there a branch already for these changes so I can adapt my future workflows already?

patrickmelix:
Any update on this?

core/basejob.py
```diff
@@ -493,7 +493,7 @@ def load_external(cls, path, settings=None, molecule=None, finalize=False):
             raise FileError('Path {} does not exist, cannot load from it.'.format(path))
 
         path = os.path.abspath(path)
-        jobname = os.path.basename(path)
+        jobname = jobname or os.path.basename(path)
```
miron-scm, line 496: > Loading jobs that have been manually run often requires renaming files to fit the PLAMS conventions. By introducing this small change, at least the folder can be named differently than the files.

> I know that this can break some things, since the job is then not in the proper place. But I think if someone finds this option they know what they are doing...

I think this would then require a comment above this line or a mention in the docstring of the function that explains why you would use this argument and that it might have unintended consequences.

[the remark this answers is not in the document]
patrickmelix: Agreed, nice suggestion!

patrickmelix: Added a docstring.

PR #99, opened 2020-11-19
Speed up unit conversions

tootea:
Unit conversions have become a major bottleneck recently as new units are added. These two simple patches bring the time to create 10000 Atoms() with coordinates in a.u. down from 6 seconds to 300 ms. Everything seems to be working exactly the same way as before (including error messages for mismatched units), with the exception of conversions from "someInvalidUnit" to "someInvalidUnit", which used to fail but now silently succeed. I hope that's not a big issue.

### DIFF
--- a/tools/units.py
+++ b/tools/units.py
@@ -145,18 +145,26 @@ class Units:
     dicts['stress'] = stress
 
 
+    # Precompute a dict mapping lowercased unit names to quantityName:conversionFactor pairs
+    quantities_for_unit = {}
+    for quantity in dicts:
+        for unit, factor in dicts[quantity].items():
+            unit = unit.lower()
+            if unit not in quantities_for_unit:
+                quantities_for_unit[unit] = {}
+            quantities_for_unit[unit][quantity] = factor
+
+
     def __init__(self):
         raise UnitsError('Instances of Units cannot be created')
 
 
     @classmethod
     def find_unit(cls, unit):
-        ret = {}
-        for quantity in cls.dicts:
-            for k in cls.dicts[quantity]:
-                if k.lower() == unit.lower():
-                    ret[quantity] = k
-        return ret
+        try:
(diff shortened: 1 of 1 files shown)


patrickmelix:
I would say this is a big issue. A conversion that succeeds without error will be a major problem for the user I would say. How should the user notice if there is no error? And typos in units are a quite expectable error in an input file I think...

BvB93:
> I would say this is a big issue.

Is it? The specific case that now always passes is if `unit1 == unit2`.
Doesn't matter if they're both bogus units or not, the conversion factor in such case is `1` by definition.

tootea:
I mean the only case that doesn't fail but used to is where you try to convert from something bogus to exactly (literally) the same thing. Thus there's nothing to do and the `convert()` method just returns the input without much thinking. Of course, this is not an issue when the source or target is hardcoded somewhere in PLAMS code, because we can be pretty sure it's not a bogus unit then.

tools/units.py
```diff
@@ -145,18 +145,26 @@ class Units:
...
-        ret = {}
-        for quantity in cls.dicts:
-            for k in cls.dicts[quantity]:
-                if k.lower() == unit.lower():
-                    ret[quantity] = k
-        return ret
+        try:
+            return cls.quantities_for_unit[unit.lower()]
+        except KeyError:
+            return {}
```
BvB93, line 167:
```suggestion
        return cls.quantities_for_unit.get(unit.lower(), {})
```
In principle you could make this a bit more compact by using `dict.get()` instead of a `try`/`except` approach.

tootea: Of course, that's way better. I think it's obvious by now that I'm not that much of a Python expert.

I have also just realized that `find_unit()` doesn't have a leading underscore, so I likely shouldn't have changed its return type. I'll fix that shortly by calling `get()` directly in `conversion_ratio()` and making a separate `find_unit()` that maintains the original signature.

PR #100, opened 2021-04-03
ORCA Interface Updates

patrickmelix:
Using ORCA now, so updating the PLAMS interface for it while I go. Will be work in progress for a while, but as always I appreciate any input in the meantime.

For now (to be extended):

- ORCA interface now loaded when doing `from scm.plams import *`
- Fix docstring links
- ORCAJob:
     - Reworked  input parser
             - did not print main section at the beginning -> total chaos when running ORCA
             - did not support %maxcore keyword layout
     - Can copy/symlink restart files to the rundir
     - Supports external files for geometry input

- ORCAResults class *filling up slowly*


patrickmelix:
Using it productively for a while now, not likely to add more soon. So this can be merged from my side.

PR #101, opened 2021-04-14
Update Vibrations Recipie

patrickmelix:
~~AMS interfaces are slightly different than before. Also needs checking if still working well with ASE.~~


interfaces/adfsuite/adf.py
```diff
@@ -69,13 +69,13 @@ def get_dipole_vector(self, unit='au'):
         raise ResultsError("'Dipole' not present in 'Properties' section of {}".format(self._kfpath()))
 
 
-    def get_gradients(self, eUnit='au', lUnit='bohr'):
-        """get_gradients(eUnit='au', lUnit='bohr')
+    def get_gradients(self, energy_unit='au', dist_unit='bohr'):
```
BvB93, line 72: Note that this name change _will_ break any existing (user) code relying on the old keyword arguments.

patrickmelix: Yeah it does... I guess the big question I have here: Will SCM remove this interface soon anyway or is it gonna stay and make it worth my while to work on it?

Will make a better commit that does not break compatibility for now.

patrickmelix: Hacky fix, let me know what you think. I really don't like it if each interface uses its own keywords, they should be standardized... But not sure how to enforce that for each interface. Perhaps write it down somewhere?

patrickmelix:
I am done with the changes I needed, they are:

- Coherent use of `energy_unit` and `dist_unit` everywhere (not `eUnit` and `lUnit`), backward compatibility in ADFResults is maintained.
- ASE atoms interface now allows passing a Settings instance to preserve charges etc when using ASE and PLAMS together.
- Vibrations recipe should now work:
     - fixed multiple bugs
     - MultiJob can now be named
     - Only writes one pickle file, not one for each displacement, as introduced a while ago in ASE
     - Removed some code duplication (kwargs in IRJob)

I tried to test as many use cases as I could think of, please let me know what I missed.

recipes/vibration.py
```diff
@@ -50,8 +51,11 @@ class VibrationsJob(MultiJob):
     """
     _result_type = VibrationsResults
 
-    def __init__(self, molecule, settings, jobType=ADFJob, get_gradients='get_gradients', reorder='inputOrder', aseVibOpt={}):
-        MultiJob.__init__(self)
+    def __init__(self, molecule, settings, name='plams.vib', jobType=ADFJob, get_gradients='get_gradients', reorder='inputOrder', aseVibOpt={}):
```
BvB93, line 54:
```suggestion
    def __init__(self, molecule, settings, jobType=ADFJob, get_gradients='get_gradients', reorder='inputOrder', aseVibOpt={}, name='plams.vib'):
```
Probably a bit safer this way, in case someone was passing `jobType` as a positional argument.

recipes/vibration.py
```diff
@@ -50,8 +51,11 @@ class VibrationsJob(MultiJob):
...
+
+        def _dummy_reorder(ignore, data):
+            return data
+        MultiJob.__init__(self, name=name)
```
BvB93, line 58:
```suggestion
        super().__init__(self, name=name)
```
Style nitpick, feel free to ignore.

recipes/vibration.py
```diff
@@ -110,8 +116,8 @@ class IRJob(VibrationsJob):
 
     * ``__init__``: get_dipole_vector, must take argument ``unit='au'``
     """
-    def __init__(self, molecule, settings, jobType=ADFJob, get_gradients='get_gradients', reorder='inputOrder', get_dipole_vector='get_dipole_vector', aseVibOpt={}):
-        VibrationsJob.__init__(self, molecule, settings, jobType=jobType, get_gradients=get_gradients, reorder=reorder, aseVibOpt=aseVibOpt)
+    def __init__(self, molecule, settings, get_dipole_vector='get_dipole_vector', **kwargs):
+        VibrationsJob.__init__(self, molecule, settings, **kwargs)
```
BvB93, line 120:
```suggestion
        super().__init__(self, molecule, settings, **kwargs)
```
Style nitpick, feel free to ignore.

recipes/vibration.py
```diff
@@ -86,15 +91,16 @@ def new_children(self):
...
-            f = [ npa(vec) for vec in self.get_grad(res, eUnit='eV', lUnit='Angstrom') ]
-            force = -1.0 * npa(self.reorder(res, f))
-            filename = osPJ(path, name+'.pckl')
-            with open(filename, 'wb') as f:
-                pickleDump(force, f, protocol=2)
+            f = [ npa(vec) for vec in self.get_grad(res, energy_unit='eV', dist_unit='Angstrom') ]
+            forces[name+'.pckl'] = -1.0 * npa(self.reorder(res, f))
+        filename = osPJ(path, self.name+'.all.pckl')
+        with open(filename, 'wb') as f:
+            pickleDump(forces, f, protocol=2)
```
BvB93, line 103: Is there a specific reason for using `protocol=2` here?
`protocol=3` has been the default since python 3.0; in python >3.8 it's been upped to `protocol=4` in fact ([docs](https://docs.python.org/3/library/pickle.html#data-stream-format)).

recipes/vibration.py
```diff
@@ -63,7 +67,7 @@ def __init__(self, molecule, settings, jobType=ADFJob, get_gradients='get_gradie
         if reorder != None:
             self.reorder = getattr(self.jobType._result_type, reorder)
         else:
-            self.reorder = lambda x: x
+            self.reorder = _dummy_reorder
```
BvB93, line 70: I'm a bit confused as to what exactly is going here.
`reorder` has a default value of `"inputOrder"`, yet I cannot find any `Results.inputOrder` attribute anywhere in PLAMS.

Secondly, why does `_dummy_reorder` have two parameters all of a suden?

patrickmelix: I checked this upon your comment. I've had this there because the ADF interface used to not have the gradients in the input order (I guess the RKFs did not have it back then). So this extra step was needed. But from what I see now, all the interfaces now by default use the correct input order, so this is indeed outdated.
I'll remove the whole thing, it is anyway better if the user overwrites the get_gradients method if needed.

And later on in the code, `self.reorder` is called with two arguments, that is why the dummy takes two. It would just fail with lambda...

patrickmelix:
@BvB93: About your `super()` suggestions, I had to revert them because they break the module. Not sure why though, seemed like a cosmetic change. But I get:
```
TypeError: __init__() got multiple values for argument 'jobType'
File: vibration.py                                                                                                                                                                                                                                                                        
Line 111: super().__init__(self, molecule, settings, **kwargs)
                                                                                                                                                                                                                                                                                          
==============Full traceback========================
File: plams                                                                                                                                                                                                                                                                               
Line 88: exec(compile(open(config.default_jobmanager.input, encoding='utf-8').read(), config.default_jobmanager.input, 'exec'), plams_namespace)
----------------------------------------------------                                                                                                                                                                                                                                      
File: input
Line 119: plamsVib = IRJob(mol, settings, name='testjobs', jobType=ORCAJob, aseVibOpt={ 'delta': 0.002, 'indices': [0] }) #, 'indices': idx                                                                                                                                               
----------------------------------------------------
File: vibration.py                                                                                                                                                                                                                                                                        
Line 111: super().__init__(self, molecule, settings, **kwargs)
----------------------------------------------------               
```
when using `super()` and all works fine if I put the actual name... :thinking:

recipes/vibration.py
patrickmelix: The reason is that it is there in ASE... Why this is the case: No clue. I'll ask there.

BvB93:
Ah, I see what's going on here, the `self` parameter shouldn't be there anymore:
``` diff
- `super().__init__(self, molecule, settings, **kwargs)`
+ `super().__init__(molecule, settings, **kwargs)`
```

recipes/vibration.py
BvB93: Seems fair; I would say that adding a comment for future reference would be a good idea though.

patrickmelix:
> Ah, I see what's going on here, the `self` parameter shouldn't be there anymore:
> 
> ```diff
> - super().__init__(self, molecule, settings, **kwargs)
> + super().__init__(molecule, settings, **kwargs)
> ```

Ouch, that is so obvious... Sorry for not seeing it. :facepalm:

recipes/vibration.py
patrickmelix: I can even give you one better:https://github.com/patrickmelix/PLAMS/tree/vibration-json
Nothing done yet, but I checked how it looks in ASE and the interface should become even easier.

BvB93 (approved):
LGTM, I'd suggest pinging mattermost so Michal or someone from SCM can get it merged.

patrickmelix:
Thx for your help!

patrickmelix:
Revised some things and added a loading functionality to auto-load jobs.

PR #109, opened 2021-07-27
Add a GitHub Actions workflow for automatic PyPi publishing whenever creating a new release

BvB93:
After setting the GitHub `PYPI_TOKEN` [secret](https://docs.github.com/en/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-an-organization), the package will now be automatically published to PyPi upon creating a release.

Note that this is only triggered for full releases ([ref](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release)), so no pre-releases.


patrickmelix:
Second that, would solve  #127.

Furthermore, how about linking to Zenodo to get a citable DOI @dormrod?

robertrueger:
We'll get to this soon. SCM has now internally switched to Git for version controlling AMS. Next step is to remove PLAMS from our main repo, hook in the PLAMS Github as a submodule instead. We then intend to change the PLAMS versioning scheme to follow the AMS versions (e.g. PLAMS 2024.103) and will start tagging and releasing new PLAMS versions.

Should happen within the next few weeks ...

dormrod (approved):
Merging and will test publishing to test.pypi.org

PR #112, opened 2021-12-01
Change `init` from a function into a context manager

BvB93:
This PR changes `plams.init` from a function into a context manager.

In addition to calling it as a normal function, as could be done previously, `plams.init` can now be used on conjunction with the `with` statement. In the latter case `plams.finish` will be automatically called upon exiting the context manager.

Examples
---------
``` python
>>> from scm.plams import Molecule, Settings, AMSJob, init, finish

>>> mol: Molecule = ...
>>> settings: Settings = ...

>>> with init():
...     job1 = AMSJob(molecule=mol, settings=settings)
...     result1 = job1.run()

# Equivalently:
>>> init()
>>> job2 = AMSJob(molecule=mol, settings=settings)
>>> result2 = job2.run()
>>> finish()

```

### DIFF
--- a/core/functions.py
+++ b/core/functions.py
@@ -62,8 +62,6 @@ class init(contextlib.AbstractContextManager):
       This function **must** be called before any other PLAMS command can be executed. Trying to do anything without it results in a crash. See also |master-script|.
     """
 
-    __slots__ = ("otherJM",)
-
     def __init__(self, path=None, folder=None, config_settings: Dict = None, quiet=False, otherJM=None, use_existing_folder=False):
         self.otherJM = otherJM
 
@@ -105,19 +103,11 @@ class init(contextlib.AbstractContextManager):
 
     def __enter__(self):
         """Enter the context manager; return |init|."""
-        return self
+        return None
 
     def __exit__(self, exc_type, exc_value, traceback):
         """Exit the context manager; call |finish|."""
         finish(self.otherJM)
-
-    def __repr__(self):
-        """Return :func:`repr(self) <repr>`."""
-        return f"<plams.init: {config.init}>"
-
-    def __bool__(self):
-        """Return :class:`bool(self) <bool>`."""
-        return config.init
 
 
 def _init_slurm():
--- a/core/functions.py
+++ b/core/functions.py
@@ -7,6 +7,7 @@ import subprocess
 import time
 import types
 import warnings
+import contextlib
 from typing import Callable, Dict, NoReturn
 
 from os.path import join as opj
@@ -24,7 +25,7 @@ config.init = False
 #===========================================================================
 
 
-def init(path=None, folder=None, config_settings:Dict=None, quiet=False, use_existing_folder=False):
+class init(contextlib.AbstractContextManager):
     """Initialize PLAMS environment. Create global ``config`` and the default |JobManager|.
 
     An empty |Settings| instance is created and populated with default settings by executing ``plams_defaults``. The following locations are used to search for the defaults file, in order of precedence:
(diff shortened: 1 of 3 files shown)


core/functions.py
```diff
@@ -36,46 +37,87 @@ def init(path=None, folder=None, config_settings:Dict=None, quiet=False, use_exi
...
-    config.init = True
+    def __enter__(self):
+        """Enter the context manager; return |init|."""
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        """Exit the context manager; call |finish|."""
+        finish(self.otherJM)
+
+    def __repr__(self):
```
miron-scm, line 114: What is the purpose of having this defined?

core/functions.py
```diff
@@ -36,46 +37,87 @@ def init(path=None, folder=None, config_settings:Dict=None, quiet=False, use_exi
...
+            import dill
+        except ImportError:
+            log('WARNING: importing dill package failed. Falling back to the default pickle module. Expect problems with pickling', 1)
+
+        config.init = True
 
-    config.init = True
+    def __enter__(self):
+        """Enter the context manager; return |init|."""
+        return self
```
miron-scm, line 108: I know this is normally the default for context managers, but I think it is better to just return `None`. There is no usecase for having an instance of this object, I think it would only be a little confusing

core/functions.py
```diff
@@ -36,46 +37,87 @@ def init(path=None, folder=None, config_settings:Dict=None, quiet=False, use_exi
...
+        >>> finish()
+
+
     .. warning::
       This function **must** be called before any other PLAMS command can be executed. Trying to do anything without it results in a crash. See also |master-script|.
     """
 
-    if config.init:
-        return
+    __slots__ = ("otherJM",)
```
miron-scm, line 65: I understand the purpose of __slots__, but since it is not used anywhere else in PLAMS I think it is a little bit confusing and I'm not sure we would be saving significant amount of memory or increasing performance this way

miron-scm:
If possible it would be nice to have a unit test proving that both the context manager syntax and the normal syntax work.

core/functions.py
BvB93: Hm, fine. The returning of an instance here was the initially the main motivator for also providing things like string representations and truthfulness, so there's little need to keep those around now (removed as well).

core/functions.py
BvB93: See https://github.com/SCM-NV/PLAMS/pull/112#discussion_r1232192760

core/functions.py
```diff
@@ -36,46 +37,87 @@ def init(path=None, folder=None, config_settings:Dict=None, quiet=False, use_exi
...
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        """Exit the context manager; call |finish|."""
+        finish(self.otherJM)
+
+    def __repr__(self):
+        """Return :func:`repr(self) <repr>`."""
+        return f"<plams.init: {config.init}>"
+
+    def __bool__(self):
```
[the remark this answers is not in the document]
BvB93: See https://github.com/SCM-NV/PLAMS/pull/112#discussion_r1232192760

core/functions.py
BvB93: I'd say it's still a good practice to be explicit with which attributes are defined for a class (which `__slots__` absolutely helps with), but it's your call. Removed as of now.

BvB93:
> If possible it would be nice to have a unit test proving that both the context manager syntax and the normal syntax work.

Sorry, but considering the PLAMS test suite (at least on github) currently lacks any sort of infrastructure for testing the likes of `init`, job managers or running actual jobs I'm going to have to decline here, as this would involve this would effectively require setting up a whole new suite of testing infrastructure.

dormrod:
A good idea and implementation, but as we are looking to deprecate `init` and `finish` (likely in favour of a different context manager to manager working directories), I am closing this PR.

PR #113, opened 2022-02-14
Allow `Molecule.as_array` to return a context manager

BvB93:
This PR adds the `context` parameter to `Molecule.as_array`, allowing the latter to return a context manager.

The purpose of this addition is to make interconverion between PLAMS molecules and NumPy arrays more convenient, 
removing the need to explicitly call both `Molecule.as_array` and `Molecule.from_array` every single time 
(which is now done automatically upon exiting the context manager).

Examples
---------

``` python
>>> from scm.plams import Molecule

>>> mol = Molecule(...)

>>> with mol.as_array(context=True) as xyz_array:
>>>     xyz_array += 5.0
>>>     xyz_array[0] = [0, 0, 0]

# Or equivalently
>>> xyz_array = mol.as_array()
>>> xyz_array += 5.0
>>> xyz_array[0] = [0, 0, 0]
>>> mol.from_array(xyz_array)
```


miron-scm:
I took inspiration from your suggestion and made some commits internally.

See now https://github.com/SCM-NV/PLAMS/blob/master/mol/molecule.py#LL2384C24-L2384C24, https://github.com/SCM-NV/PLAMS/blob/master/mol/context.py and some tests in https://github.com/SCM-NV/PLAMS/blob/master/mol/context.py.

Your example still works the same, but note you need to call:

`with mol.as_array`

Without the parentheses for it to work.

PR #121, opened 2022-09-13
Update setup.py to include all explicit dependencies

bas-rustenburg:
Add numpy/scipy supported versions to setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ setup(
     ],
     keywords         = ['molecular modeling', 'computational chemistry', 'workflow', 'python interface'],
     python_requires  = '>=3.6',
-    install_requires = ['dill>=0.2.4', 'numpy>1.23', 'scipy', 'natsort'],
+    install_requires = ['dill>=0.2.4', 'numpy', 'scipy', 'natsort'],
     packages         = packages,
     package_dir      = {'scm.plams': '.'},
     package_data     = {'scm.plams' : ['plams_defaults', 'examples/*', 'unit_tests/*', 'unit_tests/xyz/*']},
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ setup(
     ],
     keywords         = ['molecular modeling', 'computational chemistry', 'workflow', 'python interface'],
     python_requires  = '>=3.6',
-    install_requires = ['dill>=0.2.4', 'numpy==1.21', 'scipy==1.8.1', 'natsort'],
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -28,7 +28,7 @@
     ],
     keywords         = ['molecular modeling', 'computational chemistry', 'workflow', 'python interface'],
     python_requires  = '>=3.6',
-    install_requires = ['dill>=0.2.4', 'numpy', 'natsort'],
+    install_requires = ['dill>=0.2.4', 'numpy==1.21', 'scipy==1.8.1', 'natsort'],
```
BvB93, line 31: I'd strongly recommend avoiding hard versions pins like these as they can *very* easily choke dependency resolvers when multiple packages are installed.

bas-rustenburg: I guess we can try to loosen the restriction here, but I first wanted to see how to get a working version.

BvB93: What happens when you remove the version restriction? One of the issues with pinning here is that you won't find a numpy or scipy version that supports both python 3.6 and 3.10.

bas-rustenburg: I guess it's picking up a too new version of scipy for the older builds now, but restricting it further would break the higher builds.

setup.py
```diff
@@ -28,7 +28,7 @@
     ],
     keywords         = ['molecular modeling', 'computational chemistry', 'workflow', 'python interface'],
     python_requires  = '>=3.6',
-    install_requires = ['dill>=0.2.4', 'numpy', 'natsort'],
+    install_requires = ['dill>=0.2.4', 'numpy>1.23', 'scipy', 'natsort'],
```
BvB93, line 31:
```suggestion
    install_requires = ['dill>=0.2.4', 'numpy', 'scipy', 'natsort'],
```
Is there a specific reason why the versions need restricting here in the first place?

bas-rustenburg: From the point of plams, I guess it ought to be agnostic to the versions. I might need to edit the build/install script instead to work around this.

BvB93: It's strange that the previous CI run was grabbing NumPy 1.23 though. I wonder if this could be related to the use of `python setup.py install` rather than `pip install .`?

bas-rustenburg:
That looks promising now.

BvB93 (approved):
Looks good to me; chances are many users will already have `scipy` instaleld anyway and it's just in general a good package to have for scientific computations.

bas-rustenburg:
Thanks for the feedback. The import check will hopefully help to keep plams usable outside of the SCM python stack.

PR #137, opened 2024-08-26
Add visual curve to the He-He dissociation curve example SO--

mhellstr:
Add a picture with the dissociation curve to the dissociation curve example.


dormrod (approved):
Good idea, I also thought this example would benefit from this

PR #138, opened 2024-08-27
Add unit tests for SingleJob and Results classes

dormrod:
# Description

Add tests for the `SingleJob` and `Results` classes. In addition include small refactoring.

## Refactorings

There are small refactorings here. Add a `StrEnum` called `JobStatus` to explicitly define the different job status types. This is now used throughout the code. Note this is backwards compatible as string comparisons should all still work as previously.

In addition add type hints to the Job and Result classes

## Bugfix

One small bug fix flagged by tests - `Job.load` did not actually return the loaded job, now it does.

## Tests

Otherwise add tests for the `SingleJob` and `Results` classes. To facilitate this, add a `DummySingleJob` which runs a simple shell command, possibly with a wait. This allows testing of the job file generation and dependencies.

### DIFF
--- a/unit_tests/test_basejob.py
+++ b/unit_tests/test_basejob.py
@@ -2,6 +2,7 @@ import pytest
 import uuid
 from unittest.mock import patch
 from datetime import datetime
+from collections import namedtuple
 
 from scm.plams.core.settings import Settings
 from scm.plams.core.basejob import SingleJob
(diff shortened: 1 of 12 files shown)


unit_tests/test_basejob.py
```diff
@@ -0,0 +1,258 @@
...
+
+class DummySingleJob(SingleJob):
+    """
+    Single Job for testing.
+    By default, creates an input file with a unique id of the form: 'Dummy input <id>'.
+    When the job is run, creates an output file with the contents: 'Dummy output <id>'.
+    There is also an option to add an arbitrary wait period before the job runs to simulate a longer-running calculation.
+    """
+
+    def __init__(self, inp=None, cmd=None, wait=0.0, **kwargs):
```
mhellstr: add type hints

mhellstr (commented):
DummySingleJob seems instructive enough to put it somewhere with all the other job classes and not hidden away in a test, so that users can import it / so that one could create some instructive example in the docs.

dormrod:
> DummySingleJob seems instructive enough to put it somewhere with all the other job classes and not hidden away in a test, so that users can import it / so that one could create some instructive example in the docs.

Interesting idea. I don't think I would want to put it with the production jobs code, but perhaps it is worth creating a `test_utils` sub package (or similar) which contains a reusable dummy job. 

For now I think I will leave as-is in the tests, until there are enough useful components that it would be worth breaking them out and including them in the package, or actually used by an example. The class itself is very thin anyway at the moment.

unit_tests/test_basejob.py
dormrod: Added

mhellstr (approved):
Looks good. Would be nice to at some point have a pedgogical example that shows the awk_file, awk_output, grep_file, etc. similar to what is being done in the test here but more easily available to users.

PR #142, opened 2024-09-06
Add UCS Support Into AMSResults

dormrod:
# Description

The principal aim of this PR is to allow the SCM `ChemicalSystem` support within PLAMS to be extended. Currently, it is supported for a `ChemicalSystem` can be passed into an `AMSJob` in place of a `Molecule` and have the job run successfully. In addition, a `ChemicalSystem` can now be retrieved from the `AMSResults` class via analogous methods to retrieving a `Molecule`.

In addition to facilitate this, unit test coverage has been added for the ams job and results classes, along with additional type-hinting and small bug fixes, detailed below.

# Changes

### AMSJob

Unit tests added based off the output rkf files of a water geometry optimisation job. The small `ams.rkf` and `dftb.rkf` files have been added. A set of tests is then run for each of the following scenarios:
* Input molecule and input settings
* Input molecule and pisa settings
* UCS molecule and input settings

### AMSResults

Unit tests added based again on the output rkf files of the water geometry optimisation. This allows testing of a fair number of the result getter methods. 

Note these tests can only be run with an AMS installation - not on the CI on GitHub.

## AMSResults

In `AMSResults` missing type hints have been added. In addition, the following methods have been added to get UCS out:
```
def get_system(self, section: str, file: str = "ams") -> "ChemicalSystem":
def get_main_system(self) -> "ChemicalSystem":
def get_input_system(self) -> "ChemicalSystem":
```
this is in analogy with the methods to retrieve Molecule/ASE Atoms. All these methods will raise an error if scm libbase is not available. There are additional methods (such as for retrieving history), but these are the simplest starting set.

## Fixes

`ChemicalSystem` is now added to the type hint on `AMSJob`. In addition, the passed molecule is deep copied when passed into the job. This mirrors the behaviour of a `Molecule` and is generally assumed behaviour. A few small unhappy paths have also been patched in the `AMSResults` methods.

In addition, a change has been made to the `InputParserFacade`. A bug has been fixed in the non-scm version of the input parser and an efficiency improvement for the scm libbase version of the parser, which caches the instance so that the json definitions do not need to be constantly reloaded.


interfaces/adfsuite/ams.py
```diff
@@ -2193,15 +2237,20 @@ class AMSJob(SingleJob):
     """A class representing a single computation with AMS driver. The corresponding results type is |AMSResults|."""
 
     results: AMSResults
-    molecule: Union[Molecule, Dict[str, Molecule], None]
+    molecule: Optional[Union[Molecule, Dict[str, Molecule], "ChemicalSystem"]]
```
robertrueger: Isn't a ``Dict[str, "ChemicalSystem"]`` allowed here too if you want multiple System blocks in the input?

interfaces/adfsuite/ams.py
```diff
@@ -2193,15 +2237,20 @@ class AMSJob(SingleJob):
...
 
     _result_type = AMSResults
     _command = "ams"
 
-    def __init__(self, molecule: Union[Molecule, Dict[str, Molecule], None] = None, *args, **kwargs):
+    def __init__(
+        self, molecule: Optional[Union[Molecule, Dict[str, Molecule], "ChemicalSystem"]] = None, *args, **kwargs
+    ):
+        molecule = molecule.copy() if _has_scm_chemsys and isinstance(molecule, ChemicalSystem) else molecule
```
robertrueger: Same thing here: I think a ``Dict[str, "ChemicalSystem"]`` should be fine too? I'm actually wondering about the copying in case a dictionary is passed. Intuitively, I'd expect the copying to happen for all systems in the dict then, but that is not what we are currently doing.

interfaces/adfsuite/ams.py
```diff
@@ -2297,12 +2346,21 @@ def get_runscript(self):
...
 
-    def check(self):
+    def check(self) -> bool:
         """Check if ``termination status`` variable from ``General`` section of main KF file equals ``NORMAL TERMINATION``."""
         try:
             status = self.results.readrkf("General", "termination status")
+        except (FileError, KeyError) as e:
+            log(str(e), 1)
+            return False
         except:
```
robertrueger, line 2368: Unrelated to this PR, but since we are touching the method already: I think the bare except will catch things like ``KeyboardInterrupt`` and should generally be avoided.

interfaces/adfsuite/ams.py
```diff
@@ -585,14 +629,14 @@ def get_band_structure(self, bands=None, unit: str = "hartree", only_high_symmet
 
         return x, complete_spinup_data, complete_spindown_data, labels, fermi_energy
 
-    def get_engine_results(self, engine: str = None):
+    def get_engine_results(self, engine: str = None) -> Dict:
```
robertrueger: Since we are touching these lines already: `None` is not a `str`, so the correct type-hint here would be `Optional[str]`.

Lots of other places where this is wrong.

robertrueger (approved):
Some smallish comments. Looks good!

interfaces/adfsuite/ams.py
dormrod: Yes good point, updated and added a test for NEB with two system block (also for the dict of molecules case)

interfaces/adfsuite/ams.py
dormrod: Updated to now copy all the elements if a dict is passed in

interfaces/adfsuite/ams.py
dormrod: True... I am not going to change this here though, as I don't have a feel for which other errors are currently being caught by the bare except which could break the check behaviour if there is a common one

interfaces/adfsuite/ams.py
dormrod: Fixed... `mypy` is my new friend

PR #145, opened 2024-09-11
Add workflow to automatically create fix2024 PRs SO107

dormrod:
# Description

Getting simple changes into both `trunk` and `fix2024` can be slightly tedious, as currently you need to:
1. Create a branch off `trunk`
2. Make the change
3. Create a PR from the new branch into `trunk` with review and approval
4. Create a branch off `fix2024` 
5. Cherry pick the change from `trunk`
6. Make a PR into `fix2024`, with review and approval

This change proposes a solution to automate some of this process as follows:
1. The PR into `trunk` can be tagged with a label `auto-pr-fix2024`
2. Once merged, a workflow will run, create a new branch off fix, cherry-pick the commits from the PR into fix branch and create a PR
3. As this is done via a bot, the original author can approve the second PR into the fix branch

This acts to eliminate steps 4-6 in the original flow. Note that the original code still always have to be reviewed as this workflow is only triggered when merged into trunk.

# Limitations

Will not work if there are conflicts etc. But at this point a more manual merge into the fix branch would be required anyway


dormrod:
Closing this as I think this is an over-complicated solution

PR #150, opened 2024-09-19
Add missing package error and unit tests for molecule interfaces

dormrod:
# Description

The functional changes in this PR focus on the behaviour of PLAMS with optional packages / behaviour.

Outside of the required packages, some of the results management in PLAMS require optional packages like ASE/RDKit etc. which need to be installed separately. 

Previously there were a few ways this was handled:

* No handling - raise `ImportError` when the function is called. Not particularly elegant.
* Remove functions from `__all__` if it the package could not be found (rdkit).  This to me is confusing as the user would just get `ImportError: cannot import name 'from_smiles' from 'scm.plams'` which is confusing as this method is documented.

The solution proposed here is to raise a new error `MissingOptionalPackageError` when one of the results methods is called. This is added to a method through the `@requires_optional_package` decorator. This message is more informative.

In addition, add/refactor a bunch of unit tests for:

* Decorators
* Identify
* Molecule interfaces (rdkit, ase, Packmol)

N.B. can also add this in for UCS support


robertrueger (approved):
Good idea with the custom exception for missing optional packages. Much superior to what we had before.

PR #152, opened 2024-10-03
Unit tests, docstrings and minor bugfixes for molecule

dormrod:
# Description

Originally one minor bug fix for the molecule class, but extended into some more general tidy up after adding some unit tests.

## Bug Fixes

* `Molecule` charge is cast to a float when converting settings to a molecule
* `Atom` atomic number is explicitly checked to be an integer
* `Molecule` has two validation methods added `_validate_atom` and `_validate_bond`. These aim to make validation and error messages more consistent across methods on the class. These validation methods will raise a `MoleculeError` with a more helpful message
* `Molecule` method `delete_all_bonds` updated to remove the reference of the molecule from the removed bond instance

## Tests

Molecule tests have been updated to use a few different simple systems. These comprise small molecules with/without bonds and lattice information etc. which are read in from xyz files.

Each system has the relevant tests associated with it. I think for molecule it could be nice to group by system instead of method, as some methods are specific only to certain use cases.

## Other

Type hints added for methods


robertrueger (approved):
Looks good to me. Nice to have more type-hints in PLAMS in general.

PR #153, opened 2024-10-07
Add mypy check to CI pipeline

dormrod:
# Description

There were many type mismatches and potential small bugs due to incorrect type handling (particularly wrt variables which can be `None`).

This adds a check using the popular mypy package: https://mypy.readthedocs.io/en/stable/getting_started.html to make sure there are no issues arising from typing.

To get this passing for most files, for the initially most important issues, many tweaks / ignores have been added where relevant.


.github/workflows/ci.yml
```diff
@@ -101,6 +102,10 @@ jobs:
       - name: Run Flake8
         run: |
           flake8 --color never --count --config .flake8 .
+
+      - name: Run MyPy
+        run: |
+           mypy --no-warn-unused-ignores
```
robertrueger: I think this is better handled in the `mypy.ini` as `warn_unused_ignores = False`. Makes it easier for people to run mypy locally.

robertrueger (approved):
More type-hinting and checking is great for PLAMS in general, so I approve.

I'm wondering a bit how annoying it will be to have these kinds of checks fail a PR time, and if people will adopt running mypy locally first because of that. We'll see ...

.github/workflows/ci.yml
dormrod: Good point - updated.

Yes if it gets too annoying I will disable it.

PR #154, opened 2024-10-10
Bump Version to 2024.204

dormrod:
# Description

Increment version in trunk to 2024.204 to align with AMS.

Also update workflow files to match those in the fix2024 branch. Includes option to push to test PyPI and an upgrade to the latest version of `setup-python`


bas-rustenburg (approved):
LGTM

PR #155, opened 2024-10-10
Add changelog

dormrod:
# Description

Add a clearer changelog for PLAMS. Whenever a PR is made, a summary of the change can be added to this log under the `Unreleased` header. Whenever a release is made, this will be converted to a `202Y.1xx` header. This will also make it straightforward to create the "What's new in PLAMS" section of the doc at the end of the year.

I have started it by adding a list of the most significant/recent changes that will be in included in AMS 2025. I have also updated the "what's new" section of the docs to show how it might be constructed from the changelog.

Finally, the last standalone document build warnings due to missing files have been mitigated.


mhellstr (dismissed):
Nice, but I think this Changelog should be part of the docs so that it is accessible to normal users who do not go to Github. A.k.a. the "what's new" section.

Things that only concern developers (generate_example.sh, Github workflows) should ideally be indicated as such.

dormrod:
> Nice, but I think this Changelog should be part of the docs so that it is accessible to normal users who do not go to Github. A.k.a. the "what's new" section.
> 
> Things that only concern developers (generate_example.sh, Github workflows) should ideally be indicated as such.

Yes I was thinking about this, but also didn't want to "pollute" the "what's new" section with minor issues / implementation details. 

So you think the changelog should be automatically pulled into the docs? Maybe it should be formatted in a way that it can be dropped straight in a .rst file then 🤔

robertrueger:
I also think it would be good to see the ChangeLog somewhere in the documentation. The ChangeLogs for AMS itself can also be found there, with lots of obscure little bugfixes:
https://www.scm.com/doc/Documentation/ChangeLog.html

The ChangeLog does not need to be the same as the "What's new" section. I'd image the latter only contains the large, user-facing new things, while the ChangeLog can contain all kinds of technical details and small fixes.

dormrod:
I have also updated the "What's new" section for AMS2025, to show how this might be generated from the changelog

robertrueger (approved):
Looks good to me!

PR #156, opened 2024-10-14
Giulio benedini/csv logger

GiulioIlBen:
This is a simple csv logger. 

It has been integrated in the JobManager, such that it gives you a simple summary of your finished jobs.


core/jobmanager.py
```diff
@@ -42,7 +43,7 @@ class JobManager:
 
     """
 
-    def __init__(self, settings, path=None, folder=None, use_existing_folder=False):
+    def __init__(self, settings, path=None, folder=None, use_existing_folder=False, csv_logger: LoggerCSV = None):
```
dormrod: I agree that each job manager should have a separate log file. 

I find it odd that the PLAMS method `log` always writes to the same `default_jobmanager` log file, even when the job might be using a different job manager.

Does this need to be an argument though? Did you ever find it useful to pass in an existing csv logger?

core/csv_logger.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+class LoggerCSV:
+    """basic logger that mimic the behavior of plams.log with adding option of personalize the file logging location"""
+
+    def __init__(self, log_path="logfile.csv", stdout_level=3, date=True, time=True, init: bool = True):
+        self.log_path = Path(log_path)
+        self.stdout_level = stdout_level
+        self.date = date
+        self.time = time
```
dormrod: For consistency, these should come from the global config object. 

So `config.log.date`, `config.log.time` etc. 

See: https://github.com/SCM-NV/PLAMS/blob/trunk/core/settings.py#L544

core/csv_logger.py
```diff
@@ -0,0 +1,127 @@
...
+
+        Logs are printed independently to the CSV logfile and to the standard output based on configuration levels.
+        Date and/or time can be added based on configuration. All logging activity is/should thread safe.
+
+        :param message: The message to log as a dictionary.
+        :type message: dict
+        :param level: 1: important; 3: normal; 5: verbose; 7: debug, defaults to 0
+        :type level: int, optional
+        """
+        if not self.init:
```
dormrod: Is this needed here? For the main `log` function, it is checking whether init has been called (and therefore the default job manager set up) but here we don't need to wait for this, or anything else. The path has been set in the `__init__` function

core/csv_logger.py
```diff
@@ -0,0 +1,127 @@
...
+            if key not in headers:
+                headers.append(key)
+                new_headers_needed = True
+            row.append(message.get(key))
+        extra_headers_needed = new_headers_needed and (current_headers != [])
+
+        if extra_headers_needed:
+            raise NotImplementedError(
+                f"found new headers! {current_headers=} BUT {headers=}, this option is not currently supported"
+            )
```
dormrod: Right, because otherwise you would have to re-write the file with the missing `,,,`...

Could be an argument for json lines as you mentioned, as this would be more flexible, but might be less easier for users to interact with than a csv file.


For csv though, should the headers be inferred, or just set.

As in, should this be derived class called `JobCSVLogger` and you just call `log(job)` 🤔

core/csv_logger.py
```diff
@@ -0,0 +1,127 @@
...
+            if "Time" not in headers:
+                headers.insert(1 if self.date else 0, "Time")
+
+        # Log to file
+        with _filelock:
+            with open(self.log_path, mode="a", newline="") as file:
+                csv_writer = csv.writer(file)
+                if not current_headers:
+                    csv_writer.writerow(headers)
+                csv_writer.writerow(row)
```
dormrod: I would like to take the opportunity to move to the standard python logger, instead of handling the files/locking etc. ourselves. 

I will try for the default case `log` method and let you know how it goes...

unit_tests/test_csv_logger.py
```diff
@@ -0,0 +1,22 @@
+import os
+from scm.plams.core.csv_logger import LoggerCSV
+import pytest
+
+
+def test_csv_logger():
+    csv_logger = LoggerCSV(log_path=os.getcwd() + "/logfile.csv")
```
dormrod: Can use a temp file to avoid having to clean it up after the test:
```
with tempfile.NamedTemporaryFile() as temp_log_file:
    csv_logger = LoggerCSV(log_path=temp_log_file.name)
```

core/csv_logger.py
```diff
@@ -0,0 +1,127 @@
...
+        headers = current_headers.copy()
+        for key in message:
+            if key not in headers:
+                headers.append(key)
+                new_headers_needed = True
+            row.append(message.get(key))
+        extra_headers_needed = new_headers_needed and (current_headers != [])
+
+        if extra_headers_needed:
+            raise NotImplementedError(
```
dormrod: Maybe a value error?

core/jobmanager.py
GiulioIlBen: This gives you flexibility, because ideally if you want to log all the jobs in the `default_jobmanager` you can in principle provide that one as argument and therefore centralize the jobs logs in one file.

core/csv_logger.py
GiulioIlBen: true. It is a remnant due to historical reasons (at the beginning `LoggerCSV` was a function like the `log` one).

dormrod:
Really nice idea and will make people a lot happier I think!

dormrod (commented):
Summary of my changes:

* Removed old implementation
* Polished/improved the logger base class and simplified the CSV logger
* Updated the example with some explanatory text
* Added unit tests

dormrod (approved):

GiulioIlBen (commented):
The code looks good to me. I think we can merge it.

PR #157, opened 2024-10-15
Fix RKFHistoryFile to not reuse atoms

dormrod:
# Description

Couple of small fixes flagged by nightly tests, relating to additional validation added to the PLAMS `Molecule`.

The main issue was that this validation enforces that you cannot add an atom to a molecule which already belongs to another molecule. This previously led to undefined behaviour where the original molecule still has a reference to the atom, and the bonds in the atom can be in a different molecule etc.

This however flagged a few places where this was not being adhered to. In PLAMS itself, this was used in the `rkfhistoryfile`, which would throw an error on reading frames. This has now been fixed and a unit test added.

In addition:

* When creating a molecule from an input file, the charge is now converted to an integer if required (not just a float as in a recent change). Whilst not really of consequence, this caused a test failure due to a string representation comparison.
* Molecule error messages have been fixed to account for the odd case where the molecule associated with an atom/bond was not a true `Molecule` instance (seen due to an error in passing unnamed optional args in run discovery).


trajectories/rkffile.py
```diff
@@ -429,7 +429,7 @@ def read_frame(self, i, molecule=None):
             self.firsttime = False
 
         self.position = i
-        return self.coords, cell
```
dormrod: Think this should be a copy instead of a direct reference to the coordinates... otherwise when reading multiple frames the coords are changed as a side-effect of this method...

trajectories/rkfhistoryfile.py
```diff
@@ -341,10 +341,9 @@ def _read_coordinates(self, i, molecule, cell):
                 for at in reversed(molecule.atoms):
                     molecule.delete_atom(at)
                 molecule.properties = Settings()
-                for iel, el in enumerate(elements):
-                    atom = new_mol.atoms[iel]
-                    atom.bonds = []
-                    # atom = Atom(PT.get_atomic_number(el))
+                atoms = [new_mol.atoms[iel] for iel, _ in enumerate(elements)]
+                for atom in atoms:
+                    new_mol.delete_atom(atom)
```
dormrod, line 346: (this acts to delete the bonds on the atom)

trajectories/rkffile.py
dormrod: Intentional - reverted

ThePauliPrinciple (approved):
LGTM

PR #158, opened 2024-10-16
Use standard logger in log function

dormrod:
# Description

We are currently considering how to improve the PLAMS logging for a better user experience. 

This includes having logging of job summaries etc. as in https://github.com/SCM-NV/PLAMS/pull/156

As a first step though, a refactor of the existing `log` method in PLAMS. This currently had it's own logging implementation to write to stdout and a file, including handling the thread safety etc.

This change maintains the current behaviour exactly, but instead uses the standard python logger and formatter etc. This is more a best-practice change at the moment, but also provides a pattern for adding other logger types.

## Implementation

A wrapper class `Logger` has been added to `scm.plams.core.logging`. It is a thin wrapper over the standard library logger, which has an interface optimised for the current usage in PLAMS. It has a `configure` method and a `log` method which are both used by the general `log` function.

Note that the configure methods are called very frequently (every time log is called), but they only update the handlers if there is an actual change to the config. This is required as the current behaviour is to log to the global config default job manger, which can be changed dynamically in the script. So this can't be a one-time-setup type of configuration.

Instead of instantiating the `Logger` directly, a `LogManager` static class has also been added. This allows the loggers to be accessed by name and makes sure they are singletons. This is exposed to the user via a `get_logger` helper method.

## Example

An example usage is as follows:

Screenshot 2024-10-18 at 10 02 46

## Testing

Logging tests were added for `log` function, which passed before and after this change. 

Specific tests have also been added for the `Logger` class.

In addition, the scripting tests for PLAMS were run, which should flag any issues as the diffs would be incorrect

## Future Extensions

It would now be possible to add a `get_csv_logger` method which could return a csv logger based on a similar pattern to the current logger. There could also be an (abstract) base class for the different logger types.

### DIFF
--- a/core/logging.py
+++ b/core/logging.py
@@ -2,7 +2,32 @@ import logging
 import sys
 from typing import Optional
 
-__all__ = ["Logger"]
+
+from scm.plams.core.errors import FileError
+
+__all__ = ["Logger", "LoggerManager"]
+
+
+class LoggerManager:
+    """
+    Manages PLAMS logger instances.
+    The manager should not be instantiated directly, but loggers accessed through the ``get_logger`` method.
+    """
+
+    _loggers = {}
+
+    def __new__(cls, *args, **kwargs):
+        raise TypeError("LoggerManager cannot be directly instantiated.")
+
+    @classmethod
+    def get_logger(cls, name: str) -> "Logger":
+        """
+        Get a logger with the specified name.
+        If there is an existing logger with the same name this is returned, otherwise the logger is created.
+        """
+        if name not in cls._loggers:
+            cls._loggers[name] = Logger(name)
+        return cls._loggers[name]
 
 
 class Logger:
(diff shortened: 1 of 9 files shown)


core/logging.py
```diff
@@ -0,0 +1,144 @@
...
+        with self._lock:
+
+            # Initialise the stdout handler once
+            if self._stdout_handler is None:
+                self._stdout_handler = logging.StreamHandler(sys.stdout)
+                self._stdout_handler.setFormatter(self._formatter)
+                self._logger.addHandler(self._stdout_handler)
+
+            # Update the stdout handler level if required
+            if stdout_level != self._stdout_handler.level:
```
GiulioIlBen: I would sugget this correction:

`if stdout_level != 28 - self._stdout_handler.level:`

dormrod: Thanks, good spot - updated

GiulioIlBen (approved):
It looks good to me.

PR #159, opened 2024-10-17
Work-function-QE support

nfaguirrec:
Added functions to extract and plot results from work-function calculations made with QE-AMS -- SO103, SCMSUITE-9391

### DIFF
--- a/interfaces/adfsuite/ams.py
+++ b/interfaces/adfsuite/ams.py
@@ -1849,6 +1849,78 @@ class AMSResults(Results):
         z = (bin_edges[:-1] + bin_edges[1:]) / 2.0
         return z, density
 
+    def get_work_function_results(
+        self, energy_unit: str = "hartree", dist_unit: str = "bohr"
+    ) -> Tuple[np.ndarray, np.ndarray, np.ndarray, float, float, Tuple[float, float], Tuple[float, float]]:
+        """Return a tuple with the results of the work function calculation. This function only supports results from the Quantum Espresso engine.
+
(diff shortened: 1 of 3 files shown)


interfaces/adfsuite/ams.py
```diff
@@ -1849,6 +1849,78 @@ def get_density_along_axis(
         z = (bin_edges[:-1] + bin_edges[1:]) / 2.0
         return z, density
 
+    def get_work_function_results(
```
dormrod, line 1852: Is it worth it to add a small rkf file and unit test this method?

E.g. https://github.com/SCM-NV/PLAMS/blob/trunk/unit_tests/test_amsresults.py#L21

tools/plot.py
```diff
@@ -353,3 +351,132 @@ def plot_msd(job, start_time_fit_fs=None, ax=None):
     ax.set_title("MSD: Diffusion coefficient = {:.2e} m^2/s".format(diffusion_coefficient))
 
     return ax
+
+
+def plot_work_function(
```
dormrod, line 356: Does this need an example update to include this?

dormrod (commented):
Does this need an example update

interfaces/adfsuite/ams.py
nfaguirrec: There are two tests, but in AMS: $AMSHOME/examples/QE/WorkFunction/, $AMSHOME/examples/QE/WorkFunction/DipoleCorrection_WorkFunction

tools/plot.py
nfaguirrec: There are two tests, but in AMS: $AMSHOME/examples/QE/WorkFunction/, $AMSHOME/examples/QE/WorkFunction/DipoleCorrection_WorkFunction

dormrod (approved):

PR #160, opened 2024-10-21
Add manual trigger to AMSJobLogTailHandler SCMSUITE-10077 SO107

dormrod:
# Description 

There is a nightly test which regularly fails on MacOS when under load.

The option of forwarding of the AMS job logs to the stdout / plams log file seems to be dropping a number of the final log messages, causing the test to fail. 

The origin of the failure actually seems to be a delay in the event handler receiving the file modified signal. This is an issue as the observer is stopped immediately after the ams results are available, and so the final logs are not forwarded. In testing, if a sleep is added the logs eventually come through. 

However, a direct tail command through python show the logs have already been written. So, it appears it is actually the signal which is not firing.

This change then adds a manual trigger of the file after the results are available, to do a final processing of the log file.

A small unit test has been added to test the watch behaviour in general, although it does not pick up on this edge case.

### DIFF
--- a/interfaces/adfsuite/ams.py
+++ b/interfaces/adfsuite/ams.py
@@ -1,8 +1,8 @@
 import os
 from os.path import join as opj
 from typing import Dict, List, Literal, Set, Tuple, Union, Optional, TYPE_CHECKING, Any
-
 import numpy as np
+
 from scm.plams.core.basejob import SingleJob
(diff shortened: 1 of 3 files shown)


dormrod:
Tried on the Mac mini, looks to reliably fix

robertrueger (approved):
Looks like a good fix. Likely the thread goes down before the last notification comes in. Let's see what happens in our nightly tests ...

PR #162, opened 2024-10-21
Improving and testing plotting tools -- SO103

nfaguirrec:


dormrod (approved):
Tests now passing on all platforms.

Nice addition of these plotting tests!

dormrod (approved):

PR #163, opened 2024-10-21
Stop writing supercell properties to AMS input SCMSUITE-10074 SO107

dormrod:
# Description

The UCS is strict about which atomic properties it allows for molecule input. Currently in PLAMS, when `supercell` is called, additional `supercell` properties are added onto each atom. These are not used within PLAMS anywhere, but are serialised when creating the input file. With UCS this will cause annoyance as the input will no longer be valid.

This change therefore ignores this specific keyword when serialising the molecule. 

A unit test has been added to test the serialisation of a molecule created via supercell. In addition, the plams scripting tests have been run as well as the electric fields example. All pass as expected.

### DIFF
--- a/interfaces/adfsuite/ams.py
+++ b/interfaces/adfsuite/ams.py
@@ -2763,9 +2763,9 @@ class AMSJob(SingleJob):
 
         def serialize(sett, prefix=""):
             for key, val in sett.items():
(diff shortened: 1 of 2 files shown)


nfaguirrec (approved):
All good!

PR #167, opened 2024-11-03
Fix checking for molecule identity using "is"

tootea:
This is necessary to make PLAMS tests compatible with UCS's new content-comparing equality operator.

~~RFC: Shouldn't we also similarly adjust the `mol.atoms` comparisons on the immediately following lines? The unit tests work OK even in the current state, but it feels like `is not` would express the spirit of the assertion better (checking if things are two separate copies or just two references to the same object).~~ Done.

### DIFF
--- a/unit_tests/test_amsjob.py
+++ b/unit_tests/test_amsjob.py
@@ -81,8 +81,8 @@ EndEngine
 
         # When get molecule from job
         # Then job molecule is a deep copy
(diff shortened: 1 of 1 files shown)


dormrod (approved):
Looks good to me. And yes, I think if you could also just update the atom assertion to be `is not` that would be better too

PR #168, opened 2024-11-06
Adapt UCS amsjob test to use Angstrom SCMSUITE-10024 SO107

dormrod:
# Description

Small fix for a test using the UCS, which assumed coords were in Bohr (now in Angstrom)


robertrueger (approved):
Correct, now that UCS has coordinates in Angstrom!

PR #170, opened 2024-11-11
Mhellstr/amspipeerror into scm amspipe

mhellstr:
Move some exception classes and the AMSPipeWorker flatten_arrays out of PLAMS and into scm.amspipe.


__init__.py
```diff
@@ -148,37 +119,26 @@
...
-)
+from scm.plams.trajectories.rkffile import RKFTrajectoryFile, write_general_section, write_molecule_section
+from scm.plams.trajectories.rkfhistoryfile import RKFHistoryFile, molecules_to_rkf, rkf_filter_regions
 from scm.plams.trajectories.sdffile import SDFTrajectoryFile, create_sdf_string
 from scm.plams.trajectories.sdfhistoryfile import SDFHistoryFile
 from scm.plams.trajectories.trajectory import Trajectory
 from scm.plams.trajectories.trajectoryfile import TrajectoryFile
 from scm.plams.trajectories.xyzfile import XYZTrajectoryFile, create_xyz_string
 from scm.plams.trajectories.xyzhistoryfile import XYZHistoryFile
+from scm.plams.version import __version__
```
dormrod, line 141: Was this automatically reordered? (Makes a bigger diff but looks ok)

doc/source/general.rst
```diff
@@ -213,6 +213,7 @@ Deprecated
 Removed
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 * Legacy ``BANDJob``, ``DFTBJob``, ``UFFJob``, ``MOPACJob``, ``ReaxFFJob``, ``CSHessianADFJob`` and ``ADFJob`` have been removed. These were deprecated since AMS2019, and replaced by |AMSJob|.
+* Exception classes ``AMSPipeDecodeError``, ``AMSPipeError``, ``AMSPipeInvalidArgumentError``, ``AMSPipeLogicError``, ``AMSPipeRuntimeError``, ``AMSPipeUnknownArgumentError``, ``AMSPipeUnknownMethodError``, ``AMSPipeUnknownVersionError``, were moved from scm.plams to scm.amspipe.
```
dormrod, line 216: And in CHANGELOG.md please

interfaces/adfsuite/amsworker.py
```diff
@@ -1197,31 +1195,7 @@ def _check_process(self):
...
-            elif isinstance(val, collections.abc.Mapping):
-                out[key] = self._flatten_arrays(val)
-            elif isinstance(val, np.float32):  # for scalar float32
-                out[key] = np.float64(val)
-            elif isinstance(val, np.int32):
-                out[key] = int(val)
-            else:
-                out[key] = val
-        return out
+        return flatten_arrays(d)
```
dormrod, line 1197: Something still feels a but off here, in the separation of what lives in amspipe and plams e.g. having `flatten_arrays` in amspipe but not `unflatten_arrays`.

How is this currently working, in that how would a `AMSPipeRuntimeError` be raised currently is we are not importing/using `amspipe`? In that, is an `ASEPipeWorker` currently spun up and called somehow?

mhellstr: No, the ASEPipeWorker is for the opposite direction of the pipe protocol. It is not used by any PLAMS classes.

The AMSPipeRuntimeError is raised in self._call using AMSPipeError.from_message().

One could also move unflatten_arrays into scm.amspipe.utils, there is no harm in that.

__init__.py
mhellstr: Yes, apparently I had configured to use isort, which is probably good practice, but now I don't exactly know how to undo it.

interfaces/adfsuite/amsworker.py
dormrod: OK thanks I see it now. I feel like this `amspipe` dependency could be then refactored to be completely removed?

Yes you would need to duplicate the `flatten_array` logic in both `plams` and `amspipe`, but then they would be totally decoupled. Then you would also just need a plams exception which had the message status that could replace the one usage of `except AMSPipeRuntimeError as exc`?

Otherwise, doesn't this change introduce an issue similar to the `InputParser`, whereby people can't pip install plams in a different python env and use the `AMSWorker`, without copying over the `amspipe` package?

dormrod: Or are you saying AMS users can in fact also easily install scm.amspipe outside of the ams python stack this is not an issue?

mhellstr: AMS users can easily install scm.amspipe into any python env, they just pip install $AMSHOME/scripting/scm/amspipe.

Perhaps it's doable to have duplicate implementions of the errors and flatten_array, but it would also be nice to have a single implentation.

dormrod: Then in that case I think this is better. I am working on improving the error/help messages when a package is missing, so can easily just tell the user to do that

dormrod (approved):

PR #171, opened 2024-11-11
Add pyproject.toml with installable optional dependencies

dormrod:
# Description 
Adds a `pyproject.toml` file, the more "modern" way to setup a python project, which takes most of the functionality from the `setup.py`.

Also, all the optional dependencies in PLAMS have been mapped, and adds them as different optional extra installs via pip. 

Improves the error/help messaging to users when an optional package is not installed. 

Docs updated for more accessible "Getting Started" guide

# Details

## Build
The way we were building our plams package, with `python setup.py bdist_wheel` is deprecated and should not be used. This change uses the recommended `python -m build`. The information in the `setup.py` was moved into a `pyproject.toml` and made up-to-date in terms of tags, urls etc.

## Optional Dependencies
All methods which use optional dependencies should now be decorated with `requires_optional_package`. This raises the following error if a method is called which requires an optional package, but it is not installed:
```
The optional package 'rdkit is required for this PLAMS functionality, but is not available. It can be installed using the command: pip install 'plams[chem]'.  Please install and try again.
```
As can be seen, a series of extra optional packages have been defined in the `pyproject.toml`, split into different categories. The versions have not been too over-thought - the minimum version is simply set to the version in `amspython`. But where there was an issue installing the latest version of the package, the maximum version has also been set.

The exception was for `networkx`, where trivial changes were required to maintain compatibility with the latest version.

N.B. two packages, `sympy` and `promo` have not been added to any set of extras yet - they are used in the new feature https://github.com/SCM-NV/PLAMS/blob/trunk/tools/reaction.py#L224, and need to be figured out where they fit in

## Other

* Run tests also against python 3.12
* Add unit tests for `locate_rings`

## Testing
GitHub Actions were updated to use the new build method, and unit tests now use the actual .whl file and pip installs (instead of the source). This removes the need for the `requirements` files, which have also been removed.

# For Reviewers

Please also:
* Check the information in the `pyproject.toml` and workflows looks correct
* Advise if the extras categories chem, ams, analysis, docs, examples, dev are appropriate selections
* Check the docs improvements


dormrod:
@mhellstr is happy with the docs, groups etc.

Just need to check everything good from a technical standpoint within AMS

bas-rustenburg (approved):

PR #177, opened 2024-11-22
Rdkit reb

rebadam:
I added three things:
1) to_rdmol often fails, especially for the iffy compounds made with reactions discovery. I added an optional argument (presanitize) that iteratively sanitizes and update the bonds and atomic charges, untill sanitization succeeds.
2) I added a to_image function that makes a 2D image of a molecule. This uses the presanitize by default. It creates an image (svg (default), png, eps, pdf, jpeg), and optionally prints it to file. It works with our RDKit version, and when using a newer version, it uses newer features. I also works from jupyter notebooks.
3) I added a to_reaction_image function that tries to make a nice looking 2D reaction image.
I added some example molecules and a test script here.
[Example.tar.gz](https://github.com/user-attachments/files/17869800/Example.tar.gz)


interfaces/molecule/rdkit.py
```diff
@@ -108,7 +110,11 @@ def from_rdmol(rdkit_mol: "RDKitMol", confid: int = -1, properties: bool = True)
 
 @requires_optional_package("rdkit")
 def to_rdmol(
-    plams_mol: Molecule, sanitize: bool = True, properties: bool = True, assignChirality: bool = False
+    plams_mol: Molecule,
+    sanitize: bool = True,
+    properties: bool = True,
+    assignChirality: bool = False,
+    presanitize: bool = False,
```
dormrod, line 118: Add `presanitize` to doc-string with description - might be worth also mentioning that sanitize must also be `True` for this to apply

interfaces/molecule/rdkit.py
```diff
@@ -213,10 +219,11 @@ def plams_to_rd_bonds(bo):
 
     if sanitize:
         try:
-            Chem.SanitizeMol(rdmol)
+            if presanitize:
+                rdmol = _presanitize(plams_mol, rdmol)
```
dormrod, line 226: It would be nice to have add unit-test cases which use this method in `unit_tests/test_molecule_interfaces.TestRDKit` (see e.g. `test_to_rdmol_from_rdmol_roundtrip`).

These could take some input molecules that have issues, and check the output is santised as expected.

Do you have some examples for which this santization is useful?

interfaces/molecule/rdkit.py
```diff
@@ -1290,3 +1297,602 @@ def canonicalize_mol(mol, inplace=False, **kwargs):
         ret = mol.copy()
         ret.atoms = [at for _, at in sorted(zip(idx_rank, ret.atoms), reverse=True)]
         return ret
+
+
+@requires_optional_package("rdkit")
```
dormrod, line 1305: Updated as also requires optional package `pillow`, as users may not be using `amspython`. 

I have also just sorted out the optional dependencies for PLAMS to make them easier to install, so also added the pillow dependency here: https://github.com/SCM-NV/PLAMS/blob/trunk/pyproject.toml#L59

interfaces/molecule/rdkit.py
```diff
@@ -1290,3 +1297,602 @@ def canonicalize_mol(mol, inplace=False, **kwargs):
...
+def to_image(mol, remove_hydrogens=True, filename=None, format="svg", as_string=True):
```
dormrod: (Added type hints)

interfaces/molecule/rdkit.py
```diff
@@ -1290,3 +1297,602 @@ def canonicalize_mol(mol, inplace=False, **kwargs):
...
+    """
+    Convert single molecule to single image object
+
+    * ``mol`` -- PLAMS Molecule object
+    * ``remove_hydrogens`` -- Wether or not to remove the H-atoms from the image
+    * ``filename`` -- Optional: Name of image file to be created.
+    * ``format`` -- One of "svg", "png", "eps", "pdf", "jpeg"
```
dormrod: (renamed to `fmt` to avoid clash with python keyword)

interfaces/molecule/rdkit.py
```diff
@@ -1290,3 +1297,602 @@ def canonicalize_mol(mol, inplace=False, **kwargs):
...
+
+    if format not in classes.keys():
+        raise Exception("Image type %s not available." % (format))
+
+    rdmol = _rdmol_for_image(mol, remove_hydrogens)
+
+    # Draw the image
+    if classes[format.lower()] is None:
+        # With AMS version of RDKit MolsToGridImage fails for eps (because of paste)
+        img = Draw.MolToImage(rdmol, size=(200, 100))
```
dormrod: Might be nice to have the size as an option to the function?

interfaces/molecule/rdkit.py
```diff
@@ -1290,3 +1297,602 @@ def canonicalize_mol(mol, inplace=False, **kwargs):
...
+
+    # Add + and =>
+    nreactants = len(reactants)
+    img_text = add_plus_signs_svg(img_text, width, height, nmols, nreactants)
+    img_text = add_arrow_svg(img_text, width, height, nreactants)
+
+    return img_text
+
+
+def get_reaction_image_pil(reactants, products, format, width=200, height=100, as_string=True):
```
dormrod: Make these "private" (`_get_reaction_image_pil`) or add to `__init__`?

unit_tests/test_molecule_interfaces.py
```diff
@@ -230,6 +238,38 @@ def test_to_smiles_and_from_smiles_requires_rdkit_package(self, plams_mols, smil
             with pytest.raises(MissingOptionalPackageError):
                 from_smiles(smiles[0])
 
+    def test_to_image_and_get_reaction_image_can_generate_img_files(self, xyz_folder):
```
dormrod, line 267: I have converted the attached example to a unit test.

It's not too fancy, just makes sure the image files can be generated, but then they can be visually inspected too.

dormrod:
Looks like a great feature!

Just a couple of overall comments:

1. I tidied up a few little things I spotted as I reviewed this (type-hints, unit-test etc.)
2. It might also be good to add a notebook example to https://github.com/SCM-NV/PLAMS/tree/trunk/examples, or modify one of the existing ones, to show how to use this new functionality?
3. We have a new changelog which we will use at the end of the year to produce the "what's new" functionality - please could you also add a line in there? https://github.com/SCM-NV/PLAMS/blob/trunk/CHANGELOG.md

dormrod (approved):

PR #178, opened 2024-11-25
Add delete_atoms method and use bonding info for add_hatoms SO107

dormrod:
# Description 

Two small changes to `Molecule`:

* `delete_atoms`: add a method to delete multiple atoms in a single call. This uses "partial success" so that all passed atoms will attempted to be deleted, even if one of them fails.

* `add_hatoms`: use a `.in` file to amsprep instead of a `.xyz` so bonding information is respected when adding h-atoms. Also perform a guess bonds on the output, if there were bonds on the original molecule.


interfaces/adfsuite/ams.py
```diff
@@ -2355,7 +2355,11 @@ def copy_mol(mol):
         super().__init__(molecule, *args, **kwargs)
 
     def run(
-        self, jobrunner: "JobRunner" = None, jobmanager: "JobManager" = None, watch: bool = False, **kwargs
+        self,
+        jobrunner: Optional["JobRunner"] = None,
```
dormrod, line 2359: (not sure how these got past mypy...)

mol/molecule.py
```diff
@@ -3354,6 +3374,8 @@ def add_hatoms(self) -> "Molecule":
                 )
                 p.communicate()
                 retmol = self.__class__(f_out.name)
+                if any(self.bonds):
+                    retmol.guess_bonds()
```
GiulioIlBen: I do not understand why we should guess bonds by default. I would remove this line.

dormrod: Previously the behaviour was that if you do `add_hatoms` to a molecule that already has bonds, the result will have no bonds at all, as it recreates a molecule from an xyz. This change means that if the previous molecule has bonds, it will guess them. But I can remove it if you think it is unexpected here.

GiulioIlBen: ah true it makes sense.

dormrod: I have found a way to get the bonding information out of `amsprep`, so have used this instead of guess bonds

GiulioIlBen (changes requested):
I added a notebook for showing the add_hatoms and I decided to show also how to visualize molecules in the same notebook. 

It would be nice to obtain the `plt.Axes` of the molecule as visualized in the notebook, but I couldn't figure out how to do it. Since this is off-topic, I think we can leave the notebook as it is.

GiulioIlBen (approved):
For me it is good now.

PR #181, opened 2024-12-03
mhellstr/packmol with current structure

mhellstr:
@dormrod one existing issue with this branch (for example if you run examples/PackMo.py) is that it leaves a plams_workdir folder on disk.

Do you know how to safely delete this if it only contains this kind of packmol job, but definitely not delete it if it contains user jobs (or if user jobs are run after the packmol job)?

See SCMSUITE-10086 and SCMSUITE-7478 for details.


interfaces/molecule/packmol.py
```diff
@@ -748,6 +764,230 @@ def packmol_in_void(
...
+    if not job.ok():
+        raise PackMolError(
+            f"Try a lower density or a less skewed cell! Original file in {job.path} . "
+            + str(job.results.get_errormsg())
+        )
+    my_packed = job.results.get_main_system()
+    if not keepjob:
+        delete_job(job)
+
+    my_packed.remove_region("PACKMOL_thermostatted")
```
dormrod: I would suggest something like:

```
    previous_config = config.copy()
    try:
        config.job.pickle = False
        config.log.stdout = 0

        with tempfile.TemporaryDirectory() as tmp_dir:
            job_manager = JobManager(config.jobmanager, path=tmp_dir)
            job = AMSJob(settings=s, molecule=md_ucs, name="shakemd")
            job.run(job_manager=job_manager)

            if not job.ok():
                raise PackMolError(
                    f"Try a lower density or a less skewed cell! Original file in {job.path} . "
                    + str(job.results.get_errormsg())
                )
            my_packed = job.results.get_main_system()
            my_packed.remove_region("PACKMOL_thermostatted")

    finally:
        config.job.pickle = previous_config.job.pickle
        config.log.stdout = previous_config.log.stdout
```
which uses a scratch job manager that runs this job in a tmp dir that is then cleaned up.

Note that this makes the `keepjob` option obsolete. Note also that the `keepjob` is set to `True` as noted below, so that the job is currently being kept in the created workdir anyway.

If you want to have the option of still keeping the workdir, instead of a random tmp dir I would use a dir with a random name in the current directory, which you can choose to clean up (or not).

mhellstr:
Once merged I'll update the regression nightly test output

interfaces/molecule/packmol.py
mhellstr: done (except remove underscore in job_manager)

interfaces/molecule/packmol.py
```diff
@@ -299,6 +311,11 @@ def run(self):
         return output_molecule
 
 
+def sum_of_atomic_volumes(molecule: Molecule) -> float:
+    """Returns the sum of atomic volumes (calculated using vdW radii) in angstrom^3."""
+    return (4 / 3) * 3.14159 * sum(at.radius**3 for at in molecule)
```
dormrod: (updated to `np.pi` here and elsewhere)

interfaces/molecule/packmol.py
```diff
@@ -748,6 +765,233 @@ def packmol_in_void(
...
+            my_packed.remove_region("PACKMOL_thermostatted")
+
+    finally:
+        config.job.pickle = previous_config.job.pickle
+        config.log.stdout = previous_config.log.stdout
+
+    return my_packed
+
+
+@requires_optional_package("scm.libbase")
```
dormrod, line 840: This is bold, are we sure we want to limit this in such a way at the moment? None of the other packmol methods have such a restriction, and as libbase can't currently be installed outside amspython, we are limiting its usage to there currently?

interfaces/molecule/packmol.py
```diff
@@ -748,6 +765,233 @@ def packmol_in_void(
...
+    n_molecules = [1] + details["n_molecules"]
+    if system_for_packing_type == "supercell":
+        # Create a supercell that should encompass the target x/y/z lattice
+        # this is used for the initial packing of molecules to ensure there is no overlap
+        # with the original atoms
+        supercell = original_ucs.copy()
+        trafo = np.linalg.inv(original_ucs.lattice.vectors) @ np.array(target_lattice)
+        trafo = np.sign(trafo) * np.ceil(np.abs(trafo))
+        trafo = np.int_(trafo)
+        supercell.supercell_trafo(trafo)
```
dormrod, line 924: Also for example the UCS public interface is not yet stable... although I agree it is easier to do this in UCS.

If we don't have UCS available should there always be a fallback to the plams molecule?

interfaces/molecule/packmol.py
```diff
@@ -748,6 +765,233 @@ def packmol_in_void(
...
+            keepjob=True,
+        )
+
+    distorted.remove_atoms(range(len(original_ucs)))
+    distorted.map_atoms_continuous()
+    distorted.lattice = Lattice()  # so that we can add_other without having incompatible lattices
+
+    # this ensures that the original UCS is exactly preserved (including bonds etc.)
+    out_ucs = original_ucs.copy()
+    out_ucs.add_other(distorted)
```
dormrod, line 981: (could be `out_ucs = original_ucs + distorted`)

dormrod (approved):

PR #183, opened 2024-12-09
Improve Exception Handling in Pre and Post Run Methods SCMSUITE-10131 SO107

dormrod:
# Description 

The behaviour of exceptions raised in `prerun` and `postrun` methods of jobs was inconsistent and had issues, for example:

* `SingleJob` with a serial runner would raise the exception in the main thread and so kill the script entirely (not just fail a job)
* With a parallel runner, any exceptions would only get raised on the spawned thread and not be surfaced on the main thread, so there is a discrepancy in behaviour
* A multi-job which had a job which raised an exception in a child job would deadlock. This is because the parent job would never be notified that the job had finished

This PR makes this behaviour consistent, by catching any exception raised in these job steps, marking the job as complete, and storing the error message.

# Changes 

To achieve this, the following changes were made:

* The method `get_errormsg` which was already implemented on the `AMSJob` has been added to the `Job` base class with some basic implementation
* Add the decorator `_fail_on_exception` which handles exceptions for functions, marks the job and complete and notifies the parent multi-job
* Apply this decorator to `_prepare`, `_execute` and `_finalize`
* Store the exception trace in `_error_msg` which can be used by `get_errormsg` 

The following changes were also made:

* Fix some mypy issues that were flagged on the latest version 
* Convert the `Job` to an actual abstract base class instead of just raising exceptions when the methods were called

### DIFF
--- a/core/basejob.py
+++ b/core/basejob.py
@@ -4,8 +4,9 @@ import stat
 import threading
 import time
 from os.path import join as opj
-from typing import TYPE_CHECKING, Dict, Generator, Iterable, List, Optional, Union
+from typing import TYPE_CHECKING, Dict, Generator, Iterable, List, Optional, Union, Callable, Any
 from abc import ABC, abstractmethod
+import traceback
 
 from scm.plams.core.enums import JobStatus
 from scm.plams.core.errors import FileError, JobError, PlamsError, ResultsError
@@ -27,6 +28,26 @@ if TYPE_CHECKING:
     from scm.plams.core.jobrunner import JobRunner
 
 __all__ = ["SingleJob", "MultiJob"]
+
+
+def _fail_on_exception(func: Callable[["Job", Any], Any]) -> Callable[["Job", Any], Any]:
+    """Decorator to wrap a job method and mark the job as failed on any exception."""
+
+    def wrapper(self: "Job", *args, **kwargs):
+        try:
+            return func(self, *args, **kwargs)
+        except:
+            # Mark job status as failed and the results as complete
+            self.status = JobStatus.FAILED
+            self.results.finished.set()
+            self.results.done.set()
+            # Notify any parent multi-job of the failure
+            if self.parent and self in self.parent:
+                self.parent._notify()
+            # Store the exception message to be accessed from get_errormsg
+            self._error_msg = traceback.format_exc()
+
+    return wrapper
 
 
 class Job(ABC):
@@ -130,6 +151,7 @@ class Job(ABC):
         """
         if self.status != JobStatus.CREATED:
             raise JobError("Trying to run previously started job {}".format(self.name))
+        self._error_msg = None
         self.status = JobStatus.STARTED
         self._log_status(1)
 
@@ -217,6 +239,7 @@ class Job(ABC):
 
     # =======================================================================
 
+    @_fail_on_exception
     def _prepare(self, jobmanager: "JobManager") -> bool:
         """Prepare the job for execution. This method collects steps 1-7 from :ref:`job-life-cycle`. Should not be overridden. Returned value indicates if job execution should continue (|RPM| did not find this job as previously run)."""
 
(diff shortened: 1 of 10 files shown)


interfaces/adfsuite/ams.py
```diff
@@ -1802,8 +1802,8 @@ def get_density_along_axis(
 
         start_step, end_step, every, _ = self._get_integer_start_end_every_max(start_fs, end_fs, every_fs, None)
         nEntries = self.readrkf("History", "nEntries")
-        coords = np.array(self.get_history_property("Coords")).reshape(nEntries, -1, 3)
-        coords = coords[start_step:end_step:every]
+        history_coords = np.array(self.get_history_property("Coords")).reshape(nEntries, -1, 3)
```
dormrod, line 1805: (mypy complaint)

core/basejob.py
```diff
@@ -28,7 +30,27 @@
...
+    def wrapper(self: "Job", *args, **kwargs):
+        try:
+            return func(self, *args, **kwargs)
+        except:
+            # Mark job status as failed and the results as complete
+            self.status = JobStatus.FAILED
+            self.results.finished.set()  # type: ignore
+            self.results.done.set()  # type: ignore
+            # Notify any parent multi-job of the failure
+            if self.parent and self in self.parent:  # type: ignore
```
dormrod, line 45: This is the line without which the multi-job hangs indefinitely

mol/molecule.py
```diff
@@ -1509,7 +1509,7 @@ def get_complete_molecules_within_threshold(self, atom_indices, threshold: float
         D = distance_array(solvated_coords, solvated_coords)[zero_based_indices]
         less_equal = np.less_equal(D, threshold)
         within_threshold = np.any(less_equal, axis=0)
-        good_indices = [i for i, value in enumerate(within_threshold) if value]
+        good_indices = [i for i, value in enumerate(within_threshold) if value]  # type: ignore
```
dormrod, line 1512: (mypy complaint - this is always fine as the axis is specified)

core/basejob.py
```diff
@@ -28,7 +30,27 @@
...
 
 
-class Job:
+def _fail_on_exception(func):
+    """Decorator to wrap a job method and mark the job as failed on any exception."""
+
+    def wrapper(self: "Job", *args, **kwargs):
+        try:
+            return func(self, *args, **kwargs)
+        except:
```
robertrueger: Are we sure we want a bare `except` here? This will also capture things like `KeyboardInterrupt`, probably making ctrl+c somewhat unreliable. It will usually stop the script, but not if it happens to be inside of the `try` block when you press it. `SystemExit` will also be captured here ...

See: https://docs.python.org/3/library/exceptions.html#exception-hierarchy

core/basejob.py
```diff
@@ -204,6 +239,7 @@ def postrun(self) -> None:
 
     # =======================================================================
 
+    @_fail_on_exception
```
robertrueger, line 242: Are we sure we want to have the try block so far on the outside? Now it basically includes a lot of "base class work" like registering the Job with the `JobManager`. My feeling is that if something goes wrong in that generic part, it is likely a PLAMS bug and not an issue with the actual job being run. We also don't want to be so liberal with the exception catching that PLAMS bugs are just flagged as failed jobs. IMHO the job should only fail if there was an issue with the job itself.

robertrueger (commented):
I did not run tests yet, but code looks good to me. We should think a bit about which exceptions in job execution should be caught, and which ones should not. It seems that we are very liberally catching a lot now ...

core/basejob.py
dormrod: Thanks, updated to be `except Exception`.  I am not sure we can do better than that, as the user could raise any custom exception they like in the pre/post run

core/basejob.py
dormrod: It's a fair point. Could just wrap the the `prerun` and `postrun` methods, but I guess the issue here is how do we want a PLAMS bug to manifest? At the moment with a parallel job runner, an exception will not get flagged anyway to the user and the job will hang indefinitely. 

Now such a bug would flag as a failed job, and the error message would at least contain the exception message. My concern atm is that PLAMS is very fragile to deadlocking when any exception is raised

robertrueger: Alright!

robertrueger (approved):

PR #184, opened 2024-12-12
Add job status logging to job logger csv SCMSUITE-10132 SO107

dormrod:
# Description

Add a job status tracker to the base plams job, which stores the status and the timestamp at which it was changed.

Add a field to the job logging csv, which shows the timeline of these status changes. This can be useful for analysing how long jobs take and when they start running etc. which can be useful for job analysis.

An example would be:
```
2024-12-12 10:43:48 created -> 2024-12-12 10:43:48 started -> 2024-12-12 10:43:48 registered -> 2024-12-12 10:43:48 running -> 2024-12-12 10:43:51 finished -> 2024-12-12 10:43:51 successful
```

### DIFF
--- a/core/basejob.py
+++ b/core/basejob.py
@@ -2,9 +2,10 @@ import copy
 import os
 import stat
 import threading
+import datetime
 import time
 from os.path import join as opj
-from typing import TYPE_CHECKING, Dict, Generator, Iterable, List, Optional, Union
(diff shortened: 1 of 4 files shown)


GiulioIlBen (approved):
It is great and simple. Sorry for the delay.

PR #185, opened 2024-12-20
Improve Input File Generation For UCS and PISA SCMSUITE-10150 SO107

dormrod:
# Description

Efficiency improvements for generating AMS input files for `AMSJob`. This turned out to be a more complex change than expected due to the fact there are multiple ways of setting up the job:

* The input settings can either be a `Settings` object, or an input class from PISA. These settings can contain a `System` block which may have atoms / system modification keywords (like symmetrise etc.)
* The molecule can either be one or more `Molecule` instances or `ChemicalSystem`

The advantage of PISA/UCS is that these objects can be serialised directly to text, without having to go via a settings object themselves, which can be quite inefficient for large systems.

The complication comes when a system has been supplied on the settings object itself. In this case, for backwards compatibility, the system blocks must be merged. This requires serialisation via settings.

## Summary of Behaviour

* If there is no system block in the input settings and we have PISA, then we can serialise settings straight to text and add the system blocks. If the molecule is a chemical system, this can also be serialised directly to text.
* If there is a system block and we have PISA, we need to merge them. This requires converting all systems to settings and the PISA to settings, merging them, then overwriting the system block text. This was previously broken as it did not merge/deduplicate information correctly.
* If we do not have PISA, and there is no system block in the settings, then chemical systems can be serialised straight to text, and molecules as normal.
* If we do not have PISA, and there is a system block in the settings, then the chemical system must also be serialised to a settings object still

## Summary of Improvements

* In the most common case where there is no system block in the settings, chemical systems will be converted directly to text, avoiding the settings object
* Existing behaviour has been fixed in a backwards compatible way, so worst case we always serialise to settings objects and combine them
* Lots of unit tests added for generating these input files through various permutations/combinations of single/multiple molecules/chemical systems with(out) PISA

### DIFF
--- a/interfaces/adfsuite/ams.py
+++ b/interfaces/adfsuite/ams.py
@@ -2576,14 +2576,14 @@ class AMSJob(SingleJob):
         Special values can be indicated with *special* argument, which should be a dictionary having types of objects as keys and functions translating these types to strings as values.
         """
 
-        def unspec(value):
+        def unspec(value) -> str:
             """Check if *value* is one of a special types and convert it to string if it is."""
             for spec_type in special:
                 if isinstance(value, spec_type):
                     return special[spec_type](value)
             return value
 
-        def serialize(key, value, indent, end="End"):
+        def serialize(key: str, value, indent: int, end: str = "End") -> str:
             """Given a *key* and its corresponding *value* from the |Settings| instance produce a snippet of the input file representing this pair.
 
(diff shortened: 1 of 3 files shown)


robertrueger (approved):
Seems to work! The new code is not exactly pretty, but still prettier and faster than the old one was. (That part of PLAMS could really use some refactoring where this basic functionality is moved out of the AMSJob class into a common base class used by our other job classes.)

PR #190, opened 2025-01-08
Serenity interface python code and examples

Frankifisu:
This pull request includes files needed for the interface to the [Serenity](https://github.com/qcserenity) software. It includes files for the relevant classes, as well as lots of examples. Since this is a third-party software in order to run the examples one needs to install Serenity (and its numerous dependencies) first.


examples/SerenityAMS_DataTransfer_examples/C2H4_CCSD.py
```diff
@@ -0,0 +1,35 @@
...
+
+sersett.input.task.CC.system = "C2H4"
+sersett.input.task.CC.level = "CCSD"
+
+serjob = SerenityJob(settings=sersett, name="Serenity_C2H4_CCSD")
+serjob.run()
+ccsd_correction = serjob.results.get_ccsd_energy_correction()
+print(f"Serenity CCSD energy correction: {ccsd_correction} hartree")
+
+finish()
```
dormrod: These argument-less `finish` calls can now safely be removed (as in other example scripts below)

interfaces/thirdparty/serenity.py
```diff
@@ -0,0 +1,182 @@
...
+            inp += parse(item, self.settings.input[item]) + "\n"
+
+        return inp
+
+    def get_runscript(self):
+        """Returned runscript: ``serenity myinput.in |tee ser_myinput.out`` or ``serenity myinput.in``"""
+        input_file = self._filename("inp")
+        output_file = self._filename("out")
+
+        # return 'serenity {} |tee ser_{}'.format(input_file, output_file)
```
dormrod: The linter is complaining about `output_file` being unused due to the commented-out line - should this be removed? If so the docstring might also need to be updated?

interfaces/thirdparty/serenity_calculator.py
```diff
@@ -0,0 +1,30 @@
+from ase.calculators.calculator import Calculator, all_changes
+from scm.plams.interfaces.thirdparty.serenity import SerenitySettings, SerenityJob
+from scm.plams.interfaces.molecule.ase import fromASE
```
dormrod: (unused, can remove)

interfaces/thirdparty/serenity_calculator.py
```diff
@@ -0,0 +1,30 @@
...
+class SerenityCalculator(Calculator):
+    implemented_properties = ["energy"]
+
+    def __init__(self, settings: SerenitySettings):
+        super().__init__()
+        self.settings = settings
+        self.results = dict()
+
+    def calculate(self, atoms=None, properties=["energy"], system_changes=all_changes):
+        init()
```
dormrod: Think you should be able to remove this without issue

interfaces/thirdparty/serenity_calculator.py
```diff
@@ -0,0 +1,30 @@
...
+        for sys in self.settings.input.system:
+            self.settings.input.system[sys].geometry = os.path.abspath("my_system.xyz")
+        job = SerenityJob(settings=self.settings, name="serenity_job")
+        job.run()
+        self.results = dict()
+        self.results["energy"] = job.results.get_energy() * Units.convert(1.0, "hartree", "eV")
+        # delete_job(job)
+
+    def clean_exit(self):
+        finish()
```
dormrod, line 30: Ditto - can be removed

interfaces/thirdparty/serenity.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+class SerenityJob(SingleJob):
+    """A Subclass of the SingleJob class representing a computational job with Serenity."""
+
+    _result_type = SerenityResults
+
+    def _get_ready(self):
+        """If molecule is defined, xyz files are generated accordingly. If there is data to transfer between ADF and Serenity, file formats are converted."""
+        SingleJob._get_ready(self)
```
dormrod:
```suggestion
        self._get_ready()
```
(as this inherits from SingleJob)

dormrod (changes requested):
Looks great, nice to get some more 3rd party stuff into PLAMS.

Couple of other things:

* Add a quick note to this changelog: https://github.com/SCM-NV/PLAMS/blob/trunk/CHANGELOG.md this will be used to generate the "what's new" at the end of the year
* Does this need to be added to the docs, with some of the examples e.g. here: https://www.scm.com/doc.trunk/plams/interfaces/thirdparty.html ?

examples/SerenityAMS_DataTransfer_examples/C2H4_CCSD.py
Frankifisu: Interesting, I missed this was implemented, should I remove all init() and finish() commands from these scripts? Is this general or are there some cases where these argument-less functions are still needed?

interfaces/thirdparty/serenity_calculator.py
Frankifisu: Here I would keep it just for readability

examples/SerenityAMS_DataTransfer_examples/C2H4_CCSD.py
dormrod: `init()` and `finish()` can be removed always. When you want to create a directory with a name other than `plams_workdir` then `init(path="foo")` is still probably the easiest way

dormrod (approved):

PR #193, opened 2025-01-10
Giulio benedini/plot grid molecules

GiulioIlBen:
I created an interface to rdkit Draw.MolsToGridImage called plot_grid_molecules.


tools/plot.py
```diff
@@ -114,6 +128,67 @@ def plot_molecule(molecule, figsize=None, ax=None, keep_axis: bool = False, **kw
...
+    img = Draw.MolsToGridImage(
+        mols=molecules,
+        molsPerRow=molsPerRow,  # Number of molecules per row
+        subImgSize=subImgSize,  # Size of each individual image
+        legends=legends,
+        **kwargs,
+    )
+
+    if save_svg_path is not None:
+        if not isinstance(img, str):
```
dormrod, line 177: N.B. updated this as I wouldn't use `assert` in production code

dormrod:
Just remember to update the changelog also. Your image also did look low quality as per your comment, but all the tests passed using the bigger image I uploaded, so I wonder if that is to do with your config somehow?

GiulioIlBen:
> Just remember to update the changelog also. Your image also did look low quality as per your comment, but all the tests passed using the bigger image I uploaded, so I wonder if that is to do with your config somehow?

I have now updated the changelog. You are probably right, but I do not know what is the cause yet.

dormrod (approved):

PR #194, opened 2025-01-10
subprocess errors comunication with plams

GiulioIlBen:
First example on how to communicate that something with amsprep is not working. I do not know if there are other examples where this occurs, but for now I have added few lines to report that error when using the function Molecule.add_hatoms.

in this case a decorator like requires_optional_package might be useful to have?

### DIFF
--- a/mol/molecule.py
+++ b/mol/molecule.py
@@ -3393,10 +3393,10 @@ class Molecule:
                         stdout=DEVNULL,
                         stderr=PIPE,  # Redirect stderr to a pipe
                     )
-                    stdout, stderr = p.communicate()
-                    stderr_str = stderr.decode("utf-8").strip() if stderr else ""
-                    if stderr_str != "":
-                        print(f"amsprep raised: {stderr} \n run the command ${command} to get more info")
+                    _, stderr = p.communicate()
+                    if stderr:
+                        stderr_str = stderr.decode("utf-8").strip()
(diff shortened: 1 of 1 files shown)


mol/molecule.py
```diff
@@ -3373,12 +3386,17 @@ def add_hatoms(self) -> "Molecule":
...
+                        command,
                         shell=True,
                         stdout=DEVNULL,
+                        stderr=PIPE,  # Redirect stderr to a pipe
                     )
-                    p.communicate()
+                    stdout, stderr = p.communicate()
+                    stderr_str = stderr.decode("utf-8").strip() if stderr else ""
+                    if stderr_str != "":
+                        print(f"amsprep raised: {stderr} \n run the command ${command} to get more info")
```
dormrod:
```suggestion
                    _, stderr = p.communicate()
                    if stderr:
                        stderr_str = stderr.decode("utf-8").strip()
                        log(f"amsprep raised: {stderr_str} \n Run the command ${command} to get more information")
```
would use the plams log method here in place of the `print`

GiulioIlBen: to use the plams log is a good idea! 
And thanks for the correction about the `stderr`.When no error is raised `stderr==b''`. I did not know that `if b''` is equivalent to `if False`

dormrod (approved):

PR #198, opened 2025-01-15
Additional Settings Methods SCMSUITE-10162 SO107

dormrod:
# Description

First set of added/modified methods to `Settings` to help with manipulation and analysis.

Changes are related to those in branch `GiulioBenedini/plams_settings_extra` (GB)

* Methods `remove` and `difference` aliased to `-=` and `-` allow subtraction of a settings object from another. These are based on `GB.subtract`
* Method `contains_nested` added to check whether a nested key tuple is in the settings. Note that only the tuple is supported, not the dot syntax for now i.e. `('a', 'b', 'c')` not `a.b.c.`. This is because it is easier to be consistent if these key tuples are input/output to the settings
* `get_nested` now has a default argument as in `GB.get_default`, this is slightly breaking as it used to return an empty `Settings` if a key could not be found, but I think this is worth it for consistency
* Add `pop_nested` as per `GB.pop_nested`
* Add `nested_keys` and `block_keys` the latter being similar to `GB.get_blocks_path`, renamed as blocks is an AMS-specific concept
* Add `compare` method similar to `GB.compare`, but it includes the added/removed and modified items in one return dict
* Add `format_in_table` similar to `GB.format_in_table` to pretty print data in a table

See unit tests and new documentation example on building AMS input settings for example usage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ This changelog is effective from the 2025 releases.
 * Logging of AMS job error messages to stdout and logfile on job failure
 * Method `get_errormsg` enforced on the `Job` base class, with a default implementation
 * Added an interface to the Serenity program through methods such as `SerenityJob`, `SerenityResults` and `SerenitySettings`
+* Methods `Settings.contains_nested` and `Settings.pop_nested` for checking if nested keys exist in a settings object and popping them
 
 ### Changed
 * Functions for optional packages (e.g. RDKit, ASE) are available even when these packages are not installed, but will raise an `MissingOptionalPackageError` when called
@@ -41,6 +42,7 @@ This changelog is effective from the 2025 releases.
 * Use standard library logger for `log` function
 * Make `Job` class inherit from `ABC` and mark abstract methods 
 * Exceptions raised in `prerun` and `postrun` will always be caught and populate error message
+* `Settings.get_nested` takes a default argument which is returned if the nested key is not present in the settings instance
 
 ### Fixed
 * `Molecule.properties.charge` is a numeric instead of string type when loading molecule from a file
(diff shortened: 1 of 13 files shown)


tools/table_formatter.py
```diff
@@ -0,0 +1,89 @@
...
+    write_all_rows = num_rows <= max_rows or max_rows == -1
+    rows_at_start = num_rows if write_all_rows else max_rows // 2
+    rows_at_end = 0 if write_all_rows else max_rows - rows_at_start
+    width_values = [v for i, v in enumerate(values) if i < rows_at_start or i >= (num_rows - rows_at_end)]
+    col_widths = [
+        max(len(truncated_header[i]), max(len(str(row[i])) for row in width_values)) for i in range(len(keys))
+    ]
+
+    # Create the separator line
+    separator = f"+-{'-+-'.join('-' * width for width in col_widths)}-+"
```
GiulioIlBen: I realized that would be nice to format it with .md format. You can do this by chainging this line in:
`separator = f"|-{'-|-'.join('-' * width for width in col_widths)}-|"`

CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ This changelog is effective from the 2025 releases.
 * Logging of AMS job error messages to stdout and logfile on job failure
 * Method `get_errormsg` enforced on the `Job` base class, with a default implementation
 * Added an interface to the Serenity program through methods such as `SerenityJob`, `SerenityResults` and `SerenitySettings`
+* Methods `Settings.nested_keys`, `Settings.branch_keys` for accessing nested keys and `Settings.contains_nested` and `Settings.pop_nested` for checking if nested keys exist in a settings object and popping them
```
GiulioIlBen: to change with `Settings.branch_keys` -> `Settings.block_keys`

GiulioIlBen (changes requested):

tools/table_formatter.py
dormrod: Very good point - I have updated to be md compatible

CHANGELOG.md
dormrod: done

GiulioIlBen (approved):
It is good to go.

PR #199, opened 2025-01-15
Serenity

Frankifisu:
I tested it after also updating the ADF part of the trunk (I couldn't run certain tests without it) and I realized that one of the changes made introduced a bug.


interfaces/thirdparty/serenity.py
```diff
@@ -65,7 +65,7 @@ class SerenityJob(SingleJob):
 
     def _get_ready(self):
         """If molecule is defined, xyz files are generated accordingly. If there is data to transfer between ADF and Serenity, file formats are converted."""
-        self._get_ready(self)
```
dormrod, line 68: Ah yes. Probably better to do:
```
super()._get_ready()
```

Also if you like you can test this with ADF before merging. In `AMSHOME/scripting/scm/plams` you can checkout the plams branch that you are working on to run the tests against it

Frankifisu: Good idea, I just tested it and it works!

dormrod (approved):

PR #203, opened 2025-01-23
Add JobAnalysis Tool SCMSUITE-10164 SO107

dormrod:
# Description

Add very substantial PR to add the first instance of the `JobAnalysis` class, which helps with analysing the outcome of jobs, and also their results.

The docs and examples have already been added, so in terms of reviewing, it would be having a look at these and seeing if you like the API.

Also has changes to the example generation script to facilitate better rendering of example tables, and PLAMS log truncation.

# Future Work

This includes nothing around plotting, which was in Giulio's branch. In principle I think this would be quite nice, but for a second piece of work if we are going to include it.


dormrod:
Based on offline feedback on PR from @GiulioIlBen:

* Removed add/remove specific field methods and replaced with single `add_standard_field`
* Added `GyrationRadius` standard field

Further ideas for future additions:

* Discussed whether to add caching into the fetching of results, decided overhead is not so great for potential issues with stale results (as job hashing doesn't account for potential changes of state in the results)
* Whether to include/extend method to just be able to compare settings blocks between jobs (leave for now)
* Further filtering to just settings fields is possible
* Whether to have a way of extracting new jobs which can be modified and be run from the analysis

PR #206, opened 2025-01-23
SO--. Add readcoskf in plams.mol

WLChen1118:
I’ve added a readcoskf function to plams.mol. 
This function is useful for reading a molecule from a COSKF file.


mol/molecule.py
```diff
@@ -3266,6 +3266,20 @@ def readrkf(self, filename: str_type, section: str_type = "Molecule", **other):
         for bo in self.bonds:
             bo.mol = self
 
+    def readcoskf(self, filename: str, **other):
```
dormrod: Please update the `Added` section of: https://github.com/SCM-NV/PLAMS/blob/trunk/CHANGELOG.md

mol/molecule.py
```diff
@@ -3309,6 +3323,8 @@ def read(self, filename, inputformat=None, **other):
         if inputformat in self.__class__._readformat:
             if inputformat == "rkf":
                 return self.readrkf(filename, **other)
+            elif inputformat == "coskf":
```
dormrod, line 3326: Do you have an example small coskf file you can add and create a small unit test similar to this? https://github.com/SCM-NV/PLAMS/blob/trunk/unit_tests/test_molecule.py#L1064

You will probably need to add `skip_if_no_ams_installation()` so it doesn't fail on GitHub tests if so

dormrod (changes requested):
Also to get the linting to pass, you need to do:

* amspython -m pip install black
* amspython -m black -t py38 -l 120 . 

in the main directory

dormrod (approved):
Looks great!

PR #208, opened 2025-01-27
Convert JobRunner Parallelisation Attributes to Properties

dormrod:
# Description 

Improvements to the interface of the `JobRunner`. Previously, the parallelisation and level of parallelisation needed to be specified on initialisation. This is ok, but the publicly exposed attributes could also be modified having an undetermined effect on the class. For example:
```
   plams.config.default_jobrunner.parallel = True
   plams.config.default_jobrunner.maxjobs = 3
```
would use a parallel runner but not set the maximum job limit, as the underlying semaphore would not be updated.

This change therefore:
* Changes `parallel`, `maxjobs` and `maxthreads` to be properties
* Sets up the semaphores when altering the values of `maxjobs` and `maxthreads`
* Renames `semaphore` to `_job_limit` for consistency and to highlight no-one should ever access this - technically breaking
* Checks the values for maxjobs/maxthreads are 0 or >1. If maxjobs is set to 1, the process locks.

The alternative would also be just to have these as readonly properties so enforce that they can be set only on initialisation. I am on the fence either way - this would be a slightly simpler change, but would not allow this nice setting directly through the default job runner.

### DIFF
--- a/core/jobrunner.py
+++ b/core/jobrunner.py
@@ -92,7 +92,7 @@ class JobRunner(metaclass=_MetaRunner):
     A |JobRunner| instance can be passed to |run| with a keyword argument ``jobrunner``. If this argument is omitted, the instance stored in ``config.default_jobrunner`` is used.
     """
 
-    def __init__(self, parallel:bool=False, maxjobs:int=0, maxthreads:int=256):
+    def __init__(self, parallel: bool = False, maxjobs: int = 0, maxthreads: int = 256):
         # Initialize protected attributes
         self._parallel = False
         self._maxjobs = 0
@@ -170,7 +170,9 @@ class JobRunner(metaclass=_MetaRunner):
         if value < 0:
             raise ValueError(f"Value of 'maxthreads' must be 0 or greater than 1, but was {value}.")
         if value == 1:
-            raise ValueError("Value of 'maxthreads' must be 0 or greater than 1. To run in serial, set 'parallel=False'.")
+            raise ValueError(
+                "Value of 'maxthreads' must be 0 or greater than 1. To run in serial, set 'parallel=False'."
+            )
 
         # Wait for any threads using the semaphore to finish up before changing the instance
         if self._maxthreads and self._jobthread_limit:
@@ -179,7 +181,6 @@ class JobRunner(metaclass=_MetaRunner):
 
         self._maxthreads = value
         self._jobthread_limit = threading.BoundedSemaphore(value) if value else None
-
 
     def call(self, runscript, workdir, out, err, runflags):
         """call(runscript, workdir, out, err, runflags)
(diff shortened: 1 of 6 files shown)


core/jobrunner.py
```diff
@@ -92,11 +92,95 @@ class JobRunner(metaclass=_MetaRunner):
...
+        When set to ``0``, no job limit is enforced.
+
+        Note that to take effect, this requires :attr:`~scm.plams.core.jobrunner.JobRunner.parallel` to be ``True``.
+        The number of jobs may also be limited by :attr:`~scm.plams.core.jobrunner.JobRunner.maxthreads`.
+
+        When jobs are running using this, updating this value will block until its usage is complete.
+        """
+        return self._maxjobs
+
+    @maxjobs.setter
```
dormrod, line 133: N.B. as per PR description, could also just make these readonly and not have the option of changing outside of initialisation

robertrueger (approved):
The new code looks much more correct than what we had. I don't think it was very popular among users to set the attributes of the default JobRunner directly, but it sure looks like it should work.

If we made them not settable at all, then people would overwrite the default JobRunner with a new instance. If they do that while other jobs are already running, they might exceed their resource limits for a while until all the jobs using the old one are done.
The new behavior of waiting for all jobs will be more conservative, and underuse things for a while. I wonder though, if the thread that wants to change the JobRunner would have a hard time acquiring these semaphores if there are other job threads also waiting to acquire them?

dormrod:
> The new code looks much more correct than what we had. I don't think it was very popular among users to set the attributes of the default JobRunner directly, but it sure looks like it should work.
> 
> If we made them not settable at all, then people would overwrite the default JobRunner with a new instance. If they do that while other jobs are already running, they might exceed their resource limits for a while until all the jobs using the old one are done. The new behavior of waiting for all jobs will be more conservative, and underuse things for a while. I wonder though, if the thread that wants to change the JobRunner would have a hard time acquiring these semaphores if there are other job threads also waiting to acquire them?

A fair point, it could struggle to acquire all the locks if there are a lot of jobs in the queue. I could swap to a reader-writer lock type pattern.

dormrod:
> The new code looks much more correct than what we had. I don't think it was very popular among users to set the attributes of the default JobRunner directly, but it sure looks like it should work.
> 
> If we made them not settable at all, then people would overwrite the default JobRunner with a new instance. If they do that while other jobs are already running, they might exceed their resource limits for a while until all the jobs using the old one are done. The new behavior of waiting for all jobs will be more conservative, and underuse things for a while. I wonder though, if the thread that wants to change the JobRunner would have a hard time acquiring these semaphores if there are other job threads also waiting to acquire them?

I have added a custom `LimitedSemaphore` which should handle the resizing and also give priority access to the setter

PR #209, opened 2025-01-28
Make job manager workdir and job_logger lazy SO107

dormrod:
# Description

Convert the `workdir` on the `JobManager` to be a lazy property, which only creates the workdir when necessary i.e. on first access.

This avoids the manager creating unnecessary workdirs, for example when only using the manager to run `load_job`.  There should be no difference in behaviour otherwise.

To facilitate this, also make the job logger class lazy. This also means no csv file is ever created if no jobs are ever run.


unit_tests/test_jobmanager.py
```diff
@@ -0,0 +1,84 @@
...
+        assert not os.path.exists(job_manager._workdir)
+
+        # When access workdir for the first time
+        # Then workdir is created
+        assert os.path.exists(job_manager.workdir)
+        assert os.path.exists(job_manager._workdir)
+
+        # When access subsequent times
+        # Then workdir is not re-created
+        assert os.path.exists(job_manager._workdir)
```
robertrueger: This should probably access `workdir` instead of `_workdir`? I also wonder how we would know that no new directory has been created here? Wouldn't this just check that a potential newly created folder would exist?

dormrod: Good spot, have updated / changed comment

robertrueger (approved):

PR #212, opened 2025-01-31
Constraints reb

rebadam:
I added constraints handling to the AMSWorker. This entails an extra method SetConstraints. Also, the GeometryOptimization now looks for a message about the convergence. If it finds it, it just adds it to the results object.

I added this to the docs, and I made an example. To have this example work, I made a minor update to the plot_grid_molecules function (by Giullio). It now guesses bonds only if the molecule does not have bonds yet. In addition, it removes the conformer from the rdkit molecule, since this generally results in better 2D images.

I also made minor changes to a previous example, which I just included here because I did not want to make a whole separate pull request for it.


GiulioIlBen (approved):
On my side these changes improves upon previous work. So it is great!

robertrueger (dismissed):

robertrueger (approved):
Looks good to me now! :)

PR #217, opened 2025-02-05
Wei lin chen/cosmosacdhb mesp

WLChen1118:
Support Densf calculation in adfcosmorscompound.py.


WLChen1118:
There are two part that not qualified F811. Is there a better way on the use of prerun()? 
Another failure is import Union but it's not used. I can clean this later.

recipes/adfcosmorscompound.py
```diff
@@ -89,8 +94,9 @@ def __init__(
         coskf_dir: Optional[str] = None,
         preoptimization: Optional[str] = None,
         singlepoint: bool = False,
+        densf2hbc: bool = True,
```
dormrod: (technically breaking if people are using positional arguments, but probably can get away with it)

recipes/adfcosmorscompound.py
```diff
@@ -58,9 +62,10 @@ class ADFCOSMORSCompoundJob(MultiJob):
         coskf_dir  : The directory in which to place the generated .coskf file.  If nothing is specified, the file will be put in the plams directory corresponding to the job.
         preoptimization  : If None, do not preoptimize with a fast engine (then initial optimization is done with ADF). Otherwise, can be one of 'UFF', 'GAFF', 'GFNFF', 'GFN1-xTB', 'ANI-2x'. Note that you need valid licenses for ForceField or DFTB or MLPotential to use these preoptimizers.
         singlepoint (bool) :  Run a singlepoint in gasphase and with solvation to generate the .coskf file on the given Molecule. (no geometry optimization). Cannot be combined with ``preoptimization``.
+        densf2hbc (bool) : Defaults to True. Performs DENSF analysis to determine the hydrogen bond center (HBC) used in COSMOSAC-DHB-MESP.
```
dormrod, line 65: Update PLAMS changelog please to say this is now supported!

recipes/adfcosmorscompound.py
```diff
@@ -428,39 +433,81 @@ def adf_settings(
...
-            Optional information to write out in the "Compound Data" section of the .coskf file
+        if densf_path is not None:
+            HBC_xyz, HBC_atom, HBC_angle, HBC_info = parse_mesp(densf_path, coskf_path)
+            write_HBC_to_COSKF(coskf_path, HBC_xyz, HBC_atom, HBC_angle, HBC_info)
 
+        if coskf_dir is not None:
+            shutil.copy2(coskf_path, os.path.join(coskf_dir, coskf_name))
+
+    @staticmethod
+    def Update_HBC_to_COSKF(coskf: str, visulization: bool = False) -> None:
```
dormrod: (naming would be `update_` with lower case for consistency)

dormrod (changes requested):
Looks good

dormrod (approved):

PR #227, opened 2025-02-24
SO--. fix bug in view_HBC.

WLChen1118:


WLChen1118:
@dormrod Hi, do you know what might cause the failure on build-doc? It seems to be related to the app.add_directive in conf.py ?

dormrod:
> @dormrod Hi, do you know what might cause the failure on build-doc? It seems to be related to the app.add_directive in conf.py ?

For some reason it is using Sphinx v8.2.1, instead of 6.2.1. Let me take a look why

dormrod:
> > @dormrod Hi, do you know what might cause the failure on build-doc? It seems to be related to the app.add_directive in conf.py ?
> 
> For some reason it is using Sphinx v8.2.1, instead of 6.2.1. Let me take a look why

Or rather, the latest sphinx version causes this issue

PR #232, opened 2025-02-27
Rosa bulo/ionconductivity

rebadam:
This moves the feature to extract ions and ionic charges from a molecular system from examples  to recipes (while adding electron affinity and ionization potentials as parameters to the PeriodicTable class). This feature is then used in the example for the computation of ionic conductivity.


tools/periodic_table.py
```diff
@@ -22,127 +22,129 @@ class PeriodicTable:
     # Note: is_metallic and is_electronegative give "pragmatic" values for these quantities for the benefit of the
     #       bond guessing algorithm.
     data: List[List] = [
-        # [symbol, mass, radius, connectors, is_metallic, is_electronegative]
+        # [symbol, mass, radius, connectors, is_metallic, is_electronegative, electron_affinity, ionization_potential]
```
dormrod, line 25: Looks fine to me, not breaking and backwards compatible so all good

doc/source/examples/IonicConductivityFromMD.rst
```diff
@@ -5,25 +5,13 @@ Ionic conductivity from ams.rkf trajectory
 
 First, an ams.rkf trajectory with ions in the system needs to be calculated, as is done, for example, in the
 `ionic conductivity Tutorial <../../Tutorials/MolecularDynamicsAndMonteCarlo/IonicConductivity.html>`__.
```
dormrod, line 7: Also include an example rkf as a download here to make more self-contained?

rebadam: An rkf file is a bit large, so I added a PLAMS script instead, to create such an example rkf file.

PR #252, opened 2025-03-17
add ViscosityGreenKubo for benzene example SO--

mhellstr:
Also add new recipe AMSViscosityFromBinLogJob and corresponding results


recipes/md/amsmdjob.py
```diff
@@ -172,6 +174,9 @@ def __init__(
 
         mdsett.TimeStep = timestep or mdsett.TimeStep or self.default_timestep
         mdsett.Trajectory.SamplingFreq = samplingfreq or mdsett.Trajectory.SamplingFreq or self.default_samplingfreq
+        mdsett.Trajectory.EngineResultsFreq = (
+            engineresultsfreq or mdsett.Trajectory.EngineResultsFreq or self.default_engineresultsfreq
```
mhellstr: @dormrod this change is a bit risky, it means if you use plams 2025 you cannot use the amsmdjob class with ams2024 - is this a concern? Probably not?

mhellstr: ...The more I think about it, the less happy I am with this change. The reason I made this change is to make the default behavior what the user is likely to want (unlike AMS which has a default behavior the user is unlikely to want, in order to keep the equally unwanted default behavior from AMS2024).

doc/source/examples/ViscosityGreenKubo/ViscosityGreenKubo.ipynb.rst
```diff
@@ -0,0 +1,143 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code:: ipython3
+
+   benzene_box = plams.packmol(plams.from_smiles("c1ccccc1", forcefield="uff"), density=0.875, n_molecules=40)
+   plams.plot_molecule(benzene_box);
+
+.. figure:: ViscosityGreenKubo_files/ViscosityGreenKubo_3_0.png
+
+Equlibration MD simulation
```
dormrod: (typo)

doc/source/examples/examples.rst
```diff
@@ -3,11 +3,22 @@
 Examples
 ========
 
-In this section we present example PLAMS scripts covering various applications, from very simple tasks (like running the same calculation for multiple molecules) to more advanced dynamic workflows.
+The PLAMS examples use the Amsterdam Modeling Suite. Get a license or free trial from www.scm.com.
 
-The example scripts use computational engines from the Amsterdam Modeling Suite, and you will need a license to run them. Contact license@scm.com for further questions.
+.. dropdown:: Note about $AMSBIN for external Python environments
```
dormrod, line 8: 👍

recipes/md/amsmdjob.py
dormrod: I do not think this is a big concern, plams2025 is only guaranteed to be 100% compatible with ams2025 I would say.

Are you saying as `Trajectory.EngineResultsFreq` is a new keyword and therefore this is technically a breaking change when using with older AMS versions? If so just put it in the breaking changes section maybe.

The could be issues, as people in theory could pip install the latest plams to use with an old ams... but even then they could just pip install the correct plams version as they both now align?

recipes/md/trajectoryanalysis.py
```diff
@@ -165,17 +165,129 @@ def postrun(self):
             f.write(f"{D}\n")
 
 
+def viscosity_double_exponential(x, A, lam, tau1, tau2):
+    return A * (lam * (1 - np.exp(-x / tau1)) + (1 - lam) * (1 - np.exp(-x / tau2)))
+
+
+class AMSViscosityFromBinLogResults(AMSAnalysisResults):
```
dormrod, line 172: This name does not really speak to me... but I see the `BinLog` is a documented thing so fair enough

recipes/md/amsmdjob.py
dormrod: Another alternative @mhellstr, add something like the following to `interfaces/adfsuite/ams.py`:

```
    def get_ams_version():
        import subprocess
        import re

        cmd = "$AMSBIN/ams --version"
        result = subprocess.run(cmd, shell=True, capture_output=True, text=True)

        match = re.search(r"release=(\d+\.\d+)", result.stdout)
        return match.group(1) if match else None

    def check_ams_version(min_version="2025.101"):
        ams_version = get_ams_version()
        return ams_version >= min_version if ams_version else False
```

and simply omit the keyword for prev versions

PR #265, opened 2025-04-10
Add context manager for config SO107

dormrod:
## Overview

This is a backwards-compatible change to PLAMS which allows the values in the global `config` object to be overridden for a specific code context using a context manager.

The motivation for this is to decouple the config settings for jobs from the global object. This allows different jobs to have different settings for e.g. logging verbosity and the default job runner/manager. The former example is already a common requirement (see comments on refactoring) whilst the latter enables the development of a new `Workspace`/`Project` component.

## Changes

* A context var `_config` which holds the config value for a specific context, it is set by the `config_context` context manager
* A context manager `config_context` has been added which copies the global config into the contextvar and allows for modification of its values
* A function `get_config` has been added, which allows fetching of config for the current code context (either from the contextvar or the global `config` instance)
* Additional helpers `LazyWrapper` and `ContextAwareThread` have been added. The former allows the lazy job runner/manager to be used across contexts and the latter the parent thread config to be copied to the new thread. This is required when spawning new threads using the parallel job runner
* All the core plams code has been switched to use `get_config` instead of accessing `config` directly, this ensures the overrides are applied throughout the PLAMS code

## Example

See the following basic example of usage:

```
from scm.plams import config_context, get_config

with config_context() as cfg:
    cfg.log.stdout = 0
    print(f"Stdout logging inside context disabled: {get_config().log.stdout == 0}")
print(f"Stdout logging outside context disabled: {get_config().log.stdout == 0}")

# Stdout logging inside context disabled: True
# Stdout logging outside context disabled: False
```

## Notes

This change is backwards compatible in the sense that changing the global `config` is still fine and will apply everywhere as it used to. 

The only thing which will not work is if a user tries to use `config_context` with code which calls `config` directly, instead of using `get_config`.

I would not expect this to really be an end user-facing context manager, (although it could be) it is more to facilitate development of other components.


core/functions.py
```diff
@@ -38,13 +42,74 @@
...
+    context_overrides = _config_overrides.get()
+    context_config = context_overrides if context_overrides.contains_nested(key_tuple) else config
+
+    # Special case for lazy values - these must first be initialized
+    if key == "default_jobrunner" and init_lazy:
+        return context_config.default_jobrunner
+    elif key == "default_jobmanager" and init_lazy:
+        return context_config.default_jobmanager
+    else:
+        return context_config.get_nested(key_tuple, default=default)
```
dormrod: These are the core changes

interfaces/molecule/packmol.py
```diff
@@ -847,13 +847,13 @@ def _run_uff_md(
...
-        config.job.pickle = False
-        config.log.stdout = 0
+    overrides = Settings()
+    overrides.job.pickle = False
+    overrides.log.stdout = 0
+    with config_overrides(overrides):
 
         with tempfile.TemporaryDirectory() as tmp_dir:
-            job_manager = JobManager(config.jobmanager, path=tmp_dir)
+            job_manager = JobManager(get_config("jobmanager"), path=tmp_dir)
```
dormrod: Example of "real" usage of context manager

unit_tests/test_functions.py
```diff
@@ -690,3 +692,148 @@ def assert_logs(
             pattern = r"log with level \d"
 
         assert all([re.fullmatch(pattern, line) is not None for line in lines])
+
+
+class TestConfig:
```
dormrod, line 703: Main tests for these changes

core/functions.py
```diff
@@ -38,13 +42,74 @@
...
+    context_config_overrides = _config_overrides.get().copy()
+    context_config_overrides.update(overrides)
+    token = _config_overrides.set(context_config_overrides)
+    try:
+        yield
+    finally:
+        _config_overrides.reset(token)
+
+
+def get_config(
```
robertrueger: This seems to go a bit against the recent changes to have a more strongly typed `ConfigSettings` object over the generic `Settings` we had in the past. Take for example this bit of code:

```python
from scm.plams import config, get_config

config.erase_workdir = True
erase_workdir = get_config("erase_workdir")
```

`config.erase_workdir` is known to be a bool attribute of the `ConfigSettings`, so I get auto-completion and warnings when using it through the global `config` directly. When I use `get_config` I'm on my own again: I can't auto-complete the `"erase_workdir"` since it's just a string, and the returned object is type-deduced just as `Any`.

Wouldn't it be a nicer user experience to have `get_config()` not take any arguments and return just a `ConfigSettings` object?

core/functions.py
```diff
@@ -38,13 +42,74 @@
...
 # An initial call to '_init' and a final call to '_finish' are made later in this module
 config = ConfigSettings()
+_config_overrides = contextvars.ContextVar("config_overrides", default=Settings())
 # ===========================================================================
 
 _logger = get_logger("plams")
 
 
+@contextmanager
+def config_overrides(overrides: Settings):
```
robertrueger: Should `overrides` be type-hinted as `Settings` or `ConfigSettings` here?

robertrueger (commented):
Making config changes within a Python context manager is a great idea! :+1: 

One thing that I think should be addressed, is that `get_config` currently breaks the new type-hinting for the config a bit, see comments below.

I also wonder if this slightly different approach could also be made to work:
```python
from scm.plams import config, ConfigContext

config.erase_workdir = True

with ConfigContext():
    config.erase_workdir = False
    assert not config.erase_workdir
    
assert config.erase_workdir
```
I guess that would only be implementable if `config` is not just a `ConfigSettings` instance, but something like a property, that can be built when it's accessed from within the context.
Maybe it's not possible, but if I had done wishful API writing as a user, without knowing anything about the implementation of this, I think this is how I would have liked this to work ...

core/functions.py
dormrod: That's a fair point, and actually I had something similar before this commit: https://github.com/SCM-NV/PLAMS/pull/265/commits/c965ab03de44d83d2e15c547e54868f5294561aa

The gnarly issue I ran into was the lazy initialisation of the JobManager. If it is already initialised, it is fine to do a copy operation of the Settings class (which maintains the reference to the job manager) and simply return the ConfigSettings.

However, if it is not initialised, you copy a `None` which is problematic, as then multiple managers can be set up in different threads. Or you force it to initialise but that can lead to side effects with unwanted job directories hanging around in certain situations. It's a bit messy at the moment.

I will have another think on this, also in regards to your comment above, as I think this would be better.

core/functions.py
dormrod: It should be `Settings` I think as otherwise the `ConfigSettings` will default certain values and potentially override some other changed values in `config`

dormrod:
> Making config changes within a Python context manager is a great idea! 👍
> 
> One thing that I think should be addressed, is that `get_config` currently breaks the new type-hinting for the config a bit, see comments below.
> 
> I also wonder if this slightly different approach could also be made to work:
> 
> ```python
> from scm.plams import config, ConfigContext
> 
> config.erase_workdir = True
> 
> with ConfigContext():
>     config.erase_workdir = False
>     assert not config.erase_workdir
>     
> assert config.erase_workdir
> ```
> 
> I guess that would only be implementable if `config` is not just a `ConfigSettings` instance, but something like a property, that can be built when it's accessed from within the context. Maybe it's not possible, but if I had done wishful API writing as a user, without knowing anything about the implementation of this, I think this is how I would have liked this to work ...

Good idea, let me think on this a bit more. It might be hard to do in a fully backwards compatible way as there are lots of scripts which modify the config directly, and these still need to work. But there might be a trick!

robertrueger:
> Good idea, let me think on this a bit more. It might be hard to do in a fully backwards compatible way as there are lots of scripts which modify the config directly, and these still need to work. But there might be a trick!

Where do you see the issue with backwards compatibility in this approach? In all the old scripts nobody uses a `ConfigContext` (since it does not exist), so any modification of the config happens on the global one. You'd only get a different behavior if you moved a piece of code that is intended to modify the config into the `ConfigContext` context, but then it seems the new proposed behavior is actually more sensible? You could slap an old script that modifies the config in wild ways into the `ConfigContext` context block to make sure it has no chance to mess with your global config.

dormrod:
> > Good idea, let me think on this a bit more. It might be hard to do in a fully backwards compatible way as there are lots of scripts which modify the config directly, and these still need to work. But there might be a trick!
> 
> Where do you see the issue with backwards compatibility in this approach? In all the old scripts nobody uses a `ConfigContext` (since it does not exist), so any modification of the config happens on the global one. You'd only get a different behavior if you moved a piece of code that is intended to modify the config into the `ConfigContext` context, but then it seems the new proposed behavior is actually more sensible? You could slap an old script that modifies the config in wild ways into the `ConfigContext` context block to make sure it has no chance to mess with your global config.

No I agree that the `ConfigContext` would be nice and non-breaking. I meant just maintaining it such that `config.erase_workdir` can be accessed/modified directly in old scripts as expected, but usages of config within the context do something more clever and read/update the context variables. I will try and give it a go!

dormrod:
> Making config changes within a Python context manager is a great idea! 👍
> 
> One thing that I think should be addressed, is that `get_config` currently breaks the new type-hinting for the config a bit, see comments below.
> 
> I also wonder if this slightly different approach could also be made to work:
> 
> ```python
> from scm.plams import config, ConfigContext
> 
> config.erase_workdir = True
> 
> with ConfigContext():
>     config.erase_workdir = False
>     assert not config.erase_workdir
>     
> assert config.erase_workdir
> ```
> 
> I guess that would only be implementable if `config` is not just a `ConfigSettings` instance, but something like a property, that can be built when it's accessed from within the context. Maybe it's not possible, but if I had done wishful API writing as a user, without knowing anything about the implementation of this, I think this is how I would have liked this to work ...

OK, I tried a bunch of different things, and I think I have found a compromise for a nicer API.

The new usage would be:
```
config.erase_workdir = True

with config_context() as cfg:
    cfg.erase_workdir = False
    assert not cfg.erase_workdir

    # Equivalently, say in a different function/class etc.
    cfg2 = get_config()
    cfg2.erase_workdir = False
assert config.erase_workdir
```

So the context manager returns the new `ConfigSettings` instance which can be modified only in the context. In addition, `get_config` now returns the `ConfigSettings` directly. So the dot notation/type hinting should all work nicely.

Note I had to add a couple of `threading_utils` to get the lazy job runner/job manager to play nice, and also for the config context to be propagated to the new threads from the job runner

robertrueger (approved):
The new user experience is much better, I think. Also type-hints on the config work as expected now.

(The lazy initialization of the default job runner and manager the global config combined with the config contexts looks quite tricky to me. I did not properly test this.)

PR #270, opened 2025-04-29
Rosa Bulo (REB) SCMSUITE-- SO107: MDAnalysisJobs now Pisa compatible

rebadam:
- The MDAnalysisJobs should now work with Pisa instead of Settings
- It is possible to add multiple previous_jobs (in the case of time analysis, it is up to the user to ensure they are consecutive)


recipes/md/trajectoryanalysis.py
```diff
@@ -30,19 +36,109 @@ def _get_max_dt_frames(self, max_correlation_time_fs):
...
+        """
+        Generate the input file
+        """
+        self._settings_to_list()
+        if self.atom_indices and self._parent_write_atoms:
+            section = getattr(self.settings.input, self._task)
+            for entry in section:
+                found = self._has_settings_entry(entry, "Atoms")
+                if found:
+                    found = self._has_settings_entry(entry.Atoms, "Atom")
```
mhellstr: Better to combine these with an "and"

mhellstr:
In general this code is quite difficult to read, isn't it easier to just convert the PISA object to Settings once and be done with it? Similar to how RD converts from settings-to-pisa, but the other way around.

unit_tests/test_tools_plot.py
```diff
@@ -393,6 +394,54 @@ def test_plot_msd(run_calculations, rkf_tools_plot, xyz_folder):
...
+    s.input.ams.MolecularDynamics.Trajectory.SamplingFreq = 1
+    s.input.ams.MolecularDynamics.TimeStep = 0.5
+    s.input.ams.MolecularDynamics.NSteps = 200
+
+    sets = Analysis()
+    sets.MeanSquareDisplacement.UseAllValues = "Yes"
+    sets.MeanSquareDisplacement.Property = "Coords"
+    sets.MeanSquareDisplacement.Atoms.Element = "O"
+    s_msd = Settings()
+    s_msd.input = sets
```
dormrod, line 421: In the `AMSJob` the input class from PISA can either fully replace the settings object, or be under `settings.input`. Does the former also work here?

Answered my own question - yes as this happens in the `Job` base class, but maybe also cover in the test somehow? i.e. using  `AMSMSDJob(job, settings=sets)`

interfaces/adfsuite/amsanalysis.py
```diff
@@ -288,14 +298,19 @@ def _serialize_mol(self):
 
         systems = AMSJob._serialize_molecule(self)
         if len(systems) > 0:
-            self.settings.input.system = systems
+            if _has_scm_pisa and isinstance(self.settings.input, DriverBlock):
```
dormrod, line 301: More for my knowledge, but do we know why this uses `SCMJob` instead of `AMSJob`?

Would we get this all "for free" if we refactored to use `AMSJob` and/or would it cause any issues?

recipes/md/trajectoryanalysis.py
```diff
@@ -30,19 +36,109 @@ def _get_max_dt_frames(self, max_correlation_time_fs):
...
 
-        historylength = self.previous_job.results.readrkf("History", "nEntries")
-        max_dt_frames = int(max_correlation_time_fs / self.previous_job.results.get_time_step())
+        # Adjust for multiple previous_jobs
+        prevjobs = self.previous_job
+        if not isinstance(prevjobs, list):
+            prevjobs = [prevjobs]
+
+        # Read history and time step
+        # FIXME: Does not account for the presence of Range of StepSize (which the user may have provided)
```
dormrod, line 46: Fix needed here?

recipes/md/trajectoryanalysis.py
```diff
@@ -30,19 +36,109 @@ def _get_max_dt_frames(self, max_correlation_time_fs):
...
+                            entry.Atoms.Atom[i] = self.atom_indices[i]
+                    else:
+                        entry.Atoms.Atom = self.atom_indices
+        return super().get_input()
+
+    def _settings_to_list(self):
+        """
+        Convert the main settings object for this Task to a list of settings objects
+        """
+        if isinstance(self.settings.input, Settings):
```
dormrod: To Matti's point, for this class it might just be easier to use:

```
        if _has_scm_pisa and isinstance(self.settings.input, DriverBlock):
            self.settings.input = self.settings.input.to_settings()
```

or similar in the c'tor, to remove all the if/else.

Trade off would be that the user would not get the benefits of PISA when accessing the job settings...

interfaces/adfsuite/amsanalysis.py
rebadam: I am not sure if it would make sense to use AMSJob as the parent. I think the main difference is that it does not produce a .rkf file, but a .kf file. And it does not need an engine block. Intuitively I don't think inheritance from AMSJob makes sense, but it is possible that it makes practical sense.

recipes/md/trajectoryanalysis.py
rebadam: Yes, but this is a separate issue. It should be fixed, but maybe not in this branch? It is already doing much more than I planned.

recipes/md/trajectoryanalysis.py
rebadam: Yes, that is definitely an option. But it would be different from the AMSJob approach. Wouldn't it be better to make the pisa interface more compatible with the Settings object? For example, give it a get() method, and maybe __contains__? There is also a bug in Pisa for non-unique entries. If you provide a single entry first, then you cannot easily add other entries. Right now I just added work-around code, but I think it should be fixed in Pisa.

unit_tests/test_tools_plot.py
rebadam: Ah, I did not know that. Will add it.

interfaces/adfsuite/amsanalysis.py
dormrod: OK thanks! I assumed `AMSXJob` would inherit from `AMSJob` and `SCMJob` was a legacy version of the `AMSJob`, but seems like it is still useful

recipes/md/trajectoryanalysis.py
dormrod: Fair enough!

recipes/md/trajectoryanalysis.py
dormrod: Yes I think PISA should be improved. Could you please raise JIRA(s) with the examples of the bugs/methods to add?

In particular I agree the `contains` is a big omission at the moment and should behave as the `Settings` contains.

I also think that the PISA/Settings objects should be more generally interchangeable to facilitate this kind of refactor. In an ideal world you shouldn't have to make this change at all... (or at least it should be very minimal)

dormrod:
Also please update: https://github.com/SCM-NV/PLAMS/blob/trunk/CHANGELOG.md

under `Unreleased/Added`

rebadam:
I don't really understand why the above is even a conflict. David, could you maybe take a look?

dormrod (approved):

PR #275, opened 2025-05-07
Wei lin chen/py crs correction

WLChen1118:
Fix the bug due to the keyword change


dormrod:
It looks like this change was originally made off `trunk`, hence the large diff and conflicts. These changes should not be put into `fix2025`, so please just cherry pick this commit: https://github.com/SCM-NV/PLAMS/commit/1c37b55abbfdd8edf3903a4c6f904bfd4a5cac74 in a new branch to merge into fix2025

PR #282, opened 2025-05-21
Add jobs_in_directory context manager SO107

dormrod:
# Description

Adds a context manager `use_subdir` (feel free to suggest a better name). This allows all jobs which are run in the given context to be put in a subdirectory of the plams workdir, allowing the result directories to have more structure. All other behaviour should remain the same.

For multi-jobs, this change applies only to the top level multi job.

This feature is used as follows:
```
 >>> with use_subdir("GeometryOptimization"):
 >>>     with use_subdir("DFTB"):
 >>>        job1.run()
 >>>     with use_subdir("ML/M3GNet"):
 >>>        job2.run()
 >>>     print(job1.path)
 >>>     print(job2.path)
path/GeometryOptimization/DFTB/job1
path/GeometryOptimization/ML/M3GNet/job2
```

The example `BasisSetBenchmark` has been updated to use this.
The tree before was:
```
plams_workdir
├── Acetylene_DZ
├── Acetylene_DZP
├── Acetylene_QZ4P
├── Acetylene_SZ
├── Acetylene_TZ2P
├── Acetylene_TZP
├── Ethane_DZ
├── Ethane_DZP
├── Ethane_QZ4P
├── Ethane_SZ
├── Ethane_TZ2P
├── Ethane_TZP
├── Ethylene_DZ
├── Ethylene_DZP
├── Ethylene_QZ4P
├── Ethylene_SZ
├── Ethylene_TZ2P
├── Ethylene_TZP
├── Methane_DZ
├── Methane_DZP
├── Methane_QZ4P
├── Methane_SZ
├── Methane_TZ2P
└── Methane_TZP
```

and the tree after is now:
```
plams_workdir
├── Acetylene
│   ├── Acetylene_DZ
│   ├── Acetylene_DZP
│   ├── Acetylene_QZ4P
│   ├── Acetylene_SZ
│   ├── Acetylene_TZ2P
│   └── Acetylene_TZP
├── Ethane
│   ├── Ethane_DZ
│   ├── Ethane_DZP
│   ├── Ethane_QZ4P
│   ├── Ethane_SZ
│   ├── Ethane_TZ2P
│   └── Ethane_TZP
├── Ethylene
│   ├── Ethylene_DZ
│   ├── Ethylene_DZP
│   ├── Ethylene_QZ4P
│   ├── Ethylene_SZ
│   ├── Ethylene_TZ2P
│   └── Ethylene_TZP
└── Methane
    ├── Methane_DZ
    ├── Methane_DZP
    ├── Methane_QZ4P
    ├── Methane_SZ
    ├── Methane_TZ2P
    └── Methane_TZP
```

Some unit tests have also been added for this functionality.

Reviewers please let me know if there are any "gotchas" I might have missed.


core/basejob.py
```diff
@@ -345,7 +345,7 @@ def _log_status(self, level: int) -> None:
 
     def _full_name(self) -> str:
         if self.parent:
-            return "/".join([self.parent._full_name(), self.name])
+            return opj(self.parent._full_name(), self.name)
```
dormrod: (for consistency, just used for duplicate job name renaming in the manager)

core/functions.py
```diff
@@ -7,7 +7,8 @@
 import types
 from os.path import dirname, expandvars, isdir, isfile
 from os.path import join as opj
-from typing import Dict, Iterable, Optional, Generator, TYPE_CHECKING
+from pathlib import Path
```
dormrod, line 10: (Will swap to using Path throughout PLAMS in a separate PR as per https://jira.scm.com/browse/SCMSUITE-8685)

examples/BasisSetBenchmark/BasisSetBenchmark.py
mhellstr, line 1: ``ja`` (and ``average_errors``) might be unbound in BasisSetBenchmark.py near the bottom of the file.

examples/BasisSetBenchmark/BasisSetBenchmark.py
```diff
@@ -5,7 +5,7 @@
 
 import sys
 import multiprocessing
-from scm.plams import JobRunner, config, from_smiles, Settings, AMSJob, init
+from scm.plams import JobRunner, config, from_smiles, Settings, AMSJob, init, use_subdir
```
mhellstr: If use_subdir can be imported then so can JobAnalysis, so you can probably assume that JobAnalysis exists (simplifies the example).

core/functions.py
SiebeLeDe: Hi dormrod, I have been using plams in combination with pathlib.Path instances. I noticed an annoying error when integrating plams with the pathlib library as unpickling jobs goes wrong when pickling jobs on a HPC cluster (often on a Linux operating system), and then unpickling them on my own Windows machine. 

Maybe good to look into this (for example, detect Path instances in the __getstate__ dunder method)?

examples/BasisSetBenchmark/BasisSetBenchmark.py
dormrod: Thanks, updated

examples/BasisSetBenchmark/BasisSetBenchmark.py
dormrod: Updated

core/functions.py
dormrod: Do you mean when you have modified the `Job.path` from:
```
self.path: Optional[str]
```
to 
```
self.path: Optional[Union[str, os.PathLike]]
```
?

If this is changed I will make sure it works cross-platform, thanks for the heads-up!

SiebeLeDe: In essence, yes, although I keep the self.path as a string.

My problem occurs when adding additional attributes to my own implementation of a MultiJob class. Below you can see the simplified structure in which I include the pl.Path instance via the `System` dataclass (I use it for defining and computing stationary states belonging to a reaction). Note: the example below is not practical, but I hope to convey the picture. 

I could also see the issue online, for example, the comment by sophros in this [stackoverflow issue](https://stackoverflow.com/questions/57286486/i-cant-load-my-model-because-i-cant-put-a-posixpath) 

```
import pathlib as pl
from dataclasses import dataclass

from scm.plams import AMSJob, Molecule, MultiJob, Results, Settings, log


@dataclass
class System:
    name: str
    mol_file: pl.Path  # Introducing here the pl.Path which will be part of the MyJob instance when adding systems to it
    calc_settings: Settings


class MyJobResults(Results):
    def __init__(self, job: "MyJob"):
        super().__init__(job)
        self.system_results = {}

    def get_energies(self) -> dict[str, float]:
        """Get the energies of all systems in the job."""
        for job, system in zip(self.job.children, self.job.systems):
            if not job.ok():
                log(f"Job {job.name} failed. Skipping energy retrieval.")
                continue

            energy = job.get_energy()
            self.system_results[system.name] = energy

        return self.system_results


class MyJob(MultiJob):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self._result_type = MyJobResults
        self.systems: list[System] = []  # Here the coupling to pl.Path is introduced which causes the pickling issue for different operating systems ()

    def add_system_job(self, name: str, mol_file: pl.Path, calc_settings: Settings) -> None:
        system = System(name=name, mol_file=mol_file, calc_settings=calc_settings)
        self.systems.append(system)

    def prerun(self) -> None:
        """Translates the systems into AMSJobs."""
        for system in self.systems:
            job = AMSJob(name=system.name, mol=Molecule(system.mol_file), settings=system.calc_settings)
            self.children.append(job)


def main():
    s = Settings()
    s.input.AMS.task = "SinglePoint"
    s.input.ADF.XC.GGA = "BP86"

    job = MyJob(settings=s)
    job.add_system_job("system1", pl.Path("system1.xyz"), s)
    job.add_system_job("system2", pl.Path("system2.xyz"), s)
    job.run()

    results: "MyJobResults" = job.results  # type: ignore # No generic typing in Results base class
    for system_name, energy in results.get_energies().items():
        print(f"Energy for {system_name}: {energy} eV")


if __name__ == "__main__":
    main()
```

SiebeLeDe: But I believe changing self.path in jobs will give the same error as both self.path as my self.systems are included in the pickling

dormrod: OK makes sense, thanks for the example

dormrod:
Thanks to feedback from @robertrueger, I have made the following changes:

* context manager is now called `run_directory`
* on use, it yields the absolute path to the run directory, assuming the default job_manager will be used
* added property `Jobmanger.current_rundir` to get the absolute path to the run directory, for a given job manager

So the idea is that the structure is now `workdir/rundir/jobdir`, with usage:
```
 >>> with run_directory("GeometryOptimization"):
 >>>     with run_directory("DFTB"):
 >>>        job1.run()
 >>>     with run_directory("ML/M3GNet"):
 >>>        job2.run()
 >>>     print(job1.path)
 >>>     print(job2.path)
path/GeometryOptimization/DFTB/job1
path/GeometryOptimization/ML/M3GNet/job2
```

N.B. run here is an adjective not a verb... i.e. "this is the directory to run in", if it is still confusing can rename again

robertrueger (approved):

dormrod:
Third time lucky, renamed to `jobs_in_directory`:

```
         >>> with jobs_in_directory("GeometryOptimization") as go_dir:
         >>>     with jobs_in_directory("DFTB"):
         >>>        job1.run()
         >>>     with jobs_in_directory("ML/M3GNet"):
         >>>        job2.run()
         >>>     print(go_dir)
         >>>     print(job1.path)
         >>>     print(job2.path)
            path/plams_workdir/GeometryOptimization
            path/plams_workdir/GeometryOptimization/DFTB/job1
            path/plams_workdir/GeometryOptimization/ML/M3GNet/job2
```

and `JobManager.current_dir_for_jobs`

I think this reads well now and is not too confusing - @mhellstr agree?

PR #284, opened 2025-06-03
Packmol supports one None value for n_molecules SO107

dormrod:
# Description

Support another specific option in the `packmol` function, namely packing with a given `density`, `box_bounds` and `n_molecules`, where N-1 of the number of molecules is specified, and the other is `None`.

For example:
```
out = packmol([sodium, chloride, water], n_molecules=[5, 5, None], density=1.029, box_bounds=[0, 0, 0, 19, 19, 19])
```
this is particularly useful when you have a dilute solute and you want to fill up the rest of the space with the solvent up to the given density.

# Implementation

See comments for the specific implementation of this feature.

Otherwise, it was mostly cleaning up the validation and adding some edge cases.

The example has also been updated, which created a large diff sadly.

# Tests

Unit tests have been added for happy/unhappy combinations in `PackmolStructure`. In addition, a large variety of test cases have been added for `packmol` function. 

A subset of these run on GitHub (with `_return_only_details=True`), otherwise they require AMS.

Further cases can easily be added in future, if other combinations need to be validated.


interfaces/molecule/packmol.py
```diff
@@ -79,8 +79,12 @@ def __init__(
         """
         self.molecule = molecule
         if fixed:
-            assert n_molecules is None or n_molecules == 1
-            assert density is None
+            if n_molecules is not None and n_molecules != 1:
```
dormrod, line 82: (convert assert -> ValueError)

interfaces/molecule/packmol.py
```diff
@@ -148,31 +152,30 @@ def get_input_block(self, fname, tolerance):
         if self.n_molecules == 0 and not self.fixed:
             return ""
         if self.fixed:
-            ret = f"""
-            structure {fname}
```
dormrod, line 152: (remove unnecessary indentations)

interfaces/molecule/packmol.py
```diff
@@ -608,14 +630,27 @@ def packmol(
...
+
+        if not n_molecules:
+            coeffs = np.int_(np.round(xs * mass_g / np.dot(xs, masses)))
+        elif one_n_molecules_missing:
+            if len(n_molecules) == 1:
+                coeffs = np.int_(np.round(mass_g / masses))
+            else:
+                n_mols = np.array([n if n is not None else np.nan for n in n_molecules])
+                missing = np.isnan(n_mols)
+                n_mols[missing] = (mass_g - np.dot(n_mols[~missing], masses[~missing])) / masses[missing]
```
dormrod, line 672: This is basically the line of code that is the main implentation

unit_tests/test_packmol.py
```diff
@@ -0,0 +1,811 @@
...
+            molecules=[water, ammonium, chloride],
+            n_molecules=[96, 2, None],
+            density=0.38,
+            box_bounds=[0, 0, 0, 20, 20, 20],
+            expected_n_atoms=300,
+            expected_n_molecules=[96, 2, 2],
+            expected_mole_fractions=[0.96, 0.02, 0.02],
+            expected_volume=8000,
+            expected_density=0.3811881797008519,
+        ),
```
dormrod, line 738: Some test cases for this specific feature

mhellstr (approved):
While testing it struck me that one would also want to do ``n_molecules=[1, None], n_atoms=1000, density=1.0`` and have the box auto-determined. So I added that. This means that n_molecules and n_atoms are no longer mutually exclusive if one of the elements of n_molecules is None.


At least from my testing everything works as intended. Feel free to sanity-check/re-review.

PR #290, opened 2025-07-08
AMSJobResults extra methods for IR spectra data getters

GiulioIlBen:
I just added a simple method to get the main engine name.


interfaces/adfsuite/ams.py
```diff
@@ -148,6 +148,27 @@ def engine_names(self) -> List[str]:
         ret.remove("ams")
         return ret
 
+    def get_main_engine_name(self) -> str:
```
dormrod: Can we add a little unit test for this? For example similar to: https://github.com/SCM-NV/PLAMS/blob/trunk/unit_tests/test_amsresults.py#L87

interfaces/adfsuite/ams.py
```diff
@@ -1041,8 +1062,33 @@ def get_force_constants(self, engine: Optional[str] = None) -> np.ndarray:
...
+
+        The *engine* argument should be the identifier of the file you wish to read. To access a file called ``something.rkf`` you need to call this function with ``engine='something'``. The *engine* argument can be omitted if there's only one engine results file in the job folder.
+        """
+        reduced_masses = np.array(
+            self._process_engine_results(lambda x: x.read("Vibrations", "ReducedMasses"), engine)
+        )
+        return reduced_masses
+
+    def get_normal_modes(
+        self, mass_weighted_hessian_eigenvectors: Optional[bool] = False, engine: Optional[str] = None
```
dormrod: Additional keyword arguments should go at the end to avoid breaking change to method signature e.g.
```
get_normal_modes("adf")
```
would no longer work

interfaces/adfsuite/ams.py
```diff
@@ -1041,8 +1062,33 @@ def get_force_constants(self, engine: Optional[str] = None) -> np.ndarray:
...
+
+        The *engine* argument should be the identifier of the file you wish to read. To access a file called ``something.rkf`` you need to call this function with ``engine='something'``. The *engine* argument can be omitted if there's only one engine results file in the job folder.
+        """
+        pvdos = self._process_engine_results(lambda x: x.read("Vibrations", "PVDOS"), engine)
+        nNormalModes = self._process_engine_results(lambda x: x.read("Vibrations", "nNormalModes"), engine)
+        nAtoms = len(self.job.molecule)
+        pvdos = np.array(pvdos).reshape(nNormalModes, nAtoms)
+        return pvdos
+
+    def get_reduced_masses(self, engine: Optional[str] = None):
```
dormrod: Add unit tests for these added/modified methods, see for example: https://github.com/SCM-NV/PLAMS/blob/trunk/unit_tests/test_amsresults.py#L622

dormrod (changes requested):
Looks good, please just fix black formatting, update the CHANGELOG and add a couple of unit tests!

interfaces/adfsuite/ams.py
mhellstr: Reminds me of this nugget from params.machine_learning.utils:
```
def get_most_recent_engine_rkf(job: plams.AMSJob) -> str:
    # for the hybrid engine MDStepXYZ-term1-MLPotential.rkf do not end up in results.rkfs.keys()
    # but is that intended or a bug in AMSResults.collect_rkfs() ?! Explicitly check for it below just to be sure.
    sorted_rkfs = natsorted([x for x in job.results.rkfs.keys() if x != "ams" and "term" not in x])
    if len(sorted_rkfs) == 0:
        raise FileNotFoundError(
            f"Couldn't find an engine results file for job {job.name}. "
            "This probably means that the simulation crashed."
        )
    current_engine_rkf = sorted_rkfs[-1]

    return str(current_engine_rkf)
```

Should perhaps be a part of AMSResults as well.

I'd assume for MD simulations one would want the ``get_main_engine_name`` to return the latest MDStepxxx.rkf.

CHANGELOG.md
```diff
@@ -19,7 +19,10 @@ This changelog is effective from the 2025 releases.
 * `get_system_blocks_as_molecules_from_input` method to extract molecules from AMS text input
 * `JobAnalysis.add_rkf_field` method to simplify adding values from an rkf to the analysis
 * `JobAnalysis.get_settings_field_key` and `JobAnalysis.get_rkf_field_key` methods to simplify getting the correct keys for analysis fields
-* `jobs_in_directory` context manager which allows jobs to be run in a subdirectory of the job manager working directory
```
dormrod, line 22: (this should not be changed)

dormrod (approved):
Tests look good, nice changes. Just a small regression in the CHANGELOG (might have been my fault with the merge)

PR #292, opened 2025-07-25
Add view function to generate images in AMSView

dormrod:
## Description

Add a function `view` to the PLAMS plotting tools, which allows users to generate an image of a molecule (or chemical system) and display it in a Jupyter notebook.

The function uses `AMSView` to generate the image in the background, which makes these images consistent with the rest of AMS. This is particularly helpful for use in notebooks which are in the docs. The function is performant and has a range of customisation options, for instance viewing along lattice vectors or principal component axes. 

I struggled to get this working in cases where there is no screen (i.e. the cluster). As such I added a range of backends which can be used to render the images. The default is AMSview, but there is also one using ams view + xvfb (if present) which should work in the headless case. Finally ASE is also supported.

## Testing

The example Visualisation notebook has been updated to use the `view` function to render images. This includes a variety of cases and settings.

## Reviewing

Take a look at the images. Try running the notebook and generating some images. Make sure the options all make sense. If you think the images can be improved in any way, make a suggestion.


mhellstr (commented):
* saferun vs safe_system_call ? Same thing or different? At least indicate in the docstring which one to use when.

* Does the image get generated when no screen is present? (e.g. when executing remotely on cluster?) amsresults2xlsx doesn't generate image in those cases.

* Should the view() function accept all these keyword arguments or some ViewConfig object instead? Seems like one may want to customize 1 million settings. (atom colors, sizes, background color, region colors, fonts, or if we're ever able to visualize coordination polyhedra or ribbons or ...)

* The name of the branch is nglview, is there a plan to make interactive rotatable structures? What would the function be called then?

* slab.guess_bonds() calls Molecule.guess_bonds() on a periodic structure, this should be discouraged. Only do that for UCS which knows how to handle periodic bonds.

* requires_ams seems useful, but it checks the AMS version and not the PLAMS version, so is there anything new in AMSview that requires 2025.204?

* I think the PIL format is fine, but there should ideally be some example which shows how to save the picture to a file as well.


* The view angles can be set more appropriately, typically you'd want to view a structure slightly tilted from head-on, see for example the picture on https://www.scm.com/doc/Workflows/SimpleActiveLearning/PythonExamples/SALNEB/main.html#create-a-supercell   - this should either be the default or there could be some named rotations (view_angle="offset_z_left", view_angle="offset_reverse_z_right", view_angle="z", etc.).

The problem with the numbers is that it is impossible to know how to set them.

dormrod:
> * saferun vs safe_system_call ? Same thing or different? At least indicate in the docstring which one to use when.

Renamed for clarity and updated doc strings

> * Does the image get generated when no screen is present? (e.g. when executing remotely on cluster?) amsresults2xlsx doesn't generate image in those cases.

No, good catch - working on this

> * Should the view() function accept all these keyword arguments or some ViewConfig object instead? Seems like one may want to customize 1 million settings. (atom colors, sizes, background color, region colors, fonts, or if we're ever able to visualize coordination polyhedra or ribbons or ...)

Went for a dual approach, added `ViewConfig` but also a subset of keyword overrides

> * The name of the branch is nglview, is there a plan to make interactive rotatable structures? What would the function be called then?

`view_interactive`? or maybe just `view(interactive=True)`...

> * slab.guess_bonds() calls Molecule.guess_bonds() on a periodic structure, this should be discouraged. Only do that for UCS which knows how to handle periodic bonds.

Removed

> * requires_ams seems useful, but it checks the AMS version and not the PLAMS version, so is there anything new in AMSview that requires 2025.204?

Yes, the command line options

> * I think the PIL format is fine, but there should ideally be some example which shows how to save the picture to a file as well.

Would add to a proper tutorial, packmol is just an initial demo

> * The view angles can be set more appropriately, typically you'd want to view a structure slightly tilted from head-on, see for example the picture on https://www.scm.com/doc/Workflows/SimpleActiveLearning/PythonExamples/SALNEB/main.html#create-a-supercell   - this should either be the default or there could be some named rotations (view_angle="offset_z_left", view_angle="offset_reverse_z_right", view_angle="z", etc.).

Added `direction` keyword with for example `tilt_z`, `along_x` etc. in addition can choose to use the lattice vectors or cartesian axes. Some limitations in AMSview mean there is some limit on flexibility here, but I think this is a nice start?

Issue #291, opened 2025-07-21
GridRunner bug: wait for access to results does not work


GiulioIlBen:
I tried to run this code on the cluster with slurm.
```python
from scm.plams.core.jobrunner import GridRunner
from scm.plams.unit_tests.test_basejob import DummySingleJob


def test_gridrunner():
    jr = GridRunner(maxjobs=2)
    run_kwargs = dict(queue="bw", nodes=1, n=24)

    job = DummySingleJob(wait=10, cmd="echo 'ResultsA'")
    job.run(jobrunner=jr, **run_kwargs)
    if job.check():
        res = job.results.grep_output()[0]
        print(res)
        job_n = DummySingleJob(wait=0.2, cmd=f"echo {res}")
    else:
        job_n = DummySingleJob(wait=0.2, cmd="echo 'ResultsA'")
    job_n.run(jobrunner=jr, **run_kwargs)


if __name__ == "__main__":
    test_gridrunner()
```
and job.check() or the access to results do stop the execution of the main python code.

dormrod:
Thanks for raising this issue. I have reproduced with the following:

```
from scm.plams.core.jobrunner import GridRunner
from scm.plams.unit_tests.test_basejob import DummySingleJob
from scm.plams import log


def test_gridrunner():
    jr = GridRunner(maxjobs=2)
    run_kwargs = dict(queue="bw", nodes=1, n=24)
    
    job = DummySingleJob(wait=10, cmd="echo 'Results'")
    job.run(jobrunner=jr, **run_kwargs)
    if job.ok():
        res = job.results.grep_output()[0]
        log(f"Job completed with result: {res}")
    else:
        log(f"Status: {job.status}")
        log("Job ok reported False")
    
    for k, v in job._call_log.items():
        print(f"Call log for '{k}': {v}")
   
    for s in job.status_log:
        print(f"Status log: {s}")    

if __name__ == "__main__":
    test_gridrunner()
```

giving output:
```
[24.07|12:12:03] JOB plamsjob STARTED
[24.07|12:12:03] Waiting for job plamsjob to finish
[24.07|12:12:03] JOB plamsjob RUNNING
[24.07|12:12:03] JOB plamsjob FINISHED
[24.07|12:12:03] Status: finished
[24.07|12:12:03] Job ok reported False
[...]

PR #295, opened 2025-07-29
Fix for finalize when job fails during execution and GridRunner regression

dormrod:
## Description 

Issue #291 flagged that the GridRunner no longer waits correctly for results.

The reason for this was two-fold:

* Commit https://github.com/SCM-NV/PLAMS/commit/22d763346417cb8e085f38626ba488dc2fc0a541 introduced the `staticmethod` decorators on some functions used by the grid runner, but these could not be called directly as they were stored in a dictionary, raising a `TypeError: 'staticmethod' object is not callable`
* An exception in `_execute` should cause the job to fail, but in this case `_finalize` was still being called and overwriting the job code back to success

Both these issues have been fixed, and the fix will also need to be added to `fix2025`.

There appears to be an integration testing hole for the GridRunner in AMS which needs to be addressed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ This changelog is effective from the 2025 releases.
 
 ### Fixed
 * Method to guess density in `packmol_around` changed to resolve large underestimations in molecular volumes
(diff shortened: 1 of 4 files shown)


GiulioIlBen (approved):
I have tested by my side and they pass as well. For me we can go ahead.

PR #300, opened 2025-08-05
packmol: use guessed density if box bounds and no n_atoms specified

mhellstr:
Use the guessed density if no n_atoms and no n_molecules specified (but if there are box_bounds).

Particularly useful when doing packmol_around on a slab, then there is no need to explicitly think about the density but one can just use the guessed one.


dormrod (approved):
Good stuff, nice tests

PR #302, opened 2025-08-06
Add annotations to functions and methods in PLAMS

ThePauliPrinciple:
This PR adds annotations to the functions and methods in PLAMS to be better in line with modern python development workflows.
A secondary goal is to keep a list of linting/style errors encountered that could be useful to address in a future PR.

To find missing annotations, `mypy` is used on the entire library with the following changed setting:
https://github.com/SCM-NV/PLAMS/blob/cd1c0d8d6c05b640abfad71478a0d75db3885d94/mypy.ini#L22
**_Noteworthy annotation style choice proposals:_**

**imports performed exclusively for type hinting**
Sometimes classes and functions are imported for the sole purpose of type checking, for these, the following pattern is used:
https://github.com/SCM-NV/PLAMS/blob/cd1c0d8d6c05b640abfad71478a0d75db3885d94/src/scm/plams/interfaces/adfsuite/forcefieldparams.py#L10-L12
When classes are imported for type hinting, they should then be provided as a literal e.g.:
https://github.com/SCM-NV/PLAMS/blob/cd1c0d8d6c05b640abfad71478a0d75db3885d94/src/scm/plams/interfaces/adfsuite/forcefieldparams.py#L290

**kf**
`KFFile` objects use a `read` method that may return different types, this makes type checking more or less impossible.
A future project may replace `read` with type specific methods e.g. `readint`.
Since the return type depends the variable being read from the kf file, the type needs to be looked up manually.
In order to help with such a future project, variables assigned to the result of `KFFile.read` will be hinted inline during this PR:
https://github.com/SCM-NV/PLAMS/blob/cd1c0d8d6c05b640abfad71478a0d75db3885d94/src/scm/plams/interfaces/adfsuite/forcefieldparams.py#L294-L296
This allows type checking now and leaves a useful hint for what type specific read method to use in the future.

**initialized lists**
Sometimes empty lists are initialized like so:
https://github.com/SCM-NV/PLAMS/blob/cd1c0d8d6c05b640abfad71478a0d75db3885d94/src/scm/plams/interfaces/adfsuite/forcefieldparams.py#L272-L275
In this case, no type checking is performed since these are `List[Unknown]` and thus match anything (as long as it is a list).
These are addressed by inline annotation when initialized as empty.
In some cases these could be replaced by more modern coding patterns e.g. list comprehensions, but this is beyond the scope of this PR.

**`args` and `kwargs`**
Currently args and kwargs are annotated as `Any`. Perhaps this can be improved.

**`rkf` files in input**
Current `SCMJob` and `SCMResults` objects are converted to their concomitant rkf file paths when used inside `SCMJob.settings.input`. These are now annotated as `TSCMJobPath`. 
https://github.com/SCM-NV/PLAMS/blob/f3f77a5e4e1f6700ae0147f4d47d6538235b3d45/src/scm/plams/interfaces/adfsuite/scmjob.py#L393-L394

**_List of style errors:_**

- E741
- E713
- E722
- E731
- Line length for docstrings is often longer than 120 characters. Create multilines or keep as is?


dormrod:
All these suggestions look very sensible to me.

Just two further points:

* There is an arbitrary exclusion list for some of the plams files: https://github.com/SCM-NV/PLAMS/blob/trunk/mypy.ini#L6 as initially only the "core" code was addressed - can any of this be easily removed?
* Can some of the other disabling checks be removed, which are related to function type annotations? Specifically I am thinking of `disallow_incomplete_defs` (so functions with only some of the arguments type-hinted) and `warn_return_any` (so functions returning `Any` when the definition has a specified return type)

ThePauliPrinciple:
> * `disallow_incomplete_defs`

I modified mypy.ini to also disallow incomplete defs and warn on return of `Any`. In the file, these are now commented out, rather than set to True, since this is closer to the intent (`strict=True`)

I will try adding the excluded parts of PLAMS after the current missing annotations are added and see if it is practical to annotate those files.

src/scm/plams/interfaces/adfsuite/amsanalysis.py
```diff
@@ -61,14 +66,14 @@ def read_data(self, kf, sec):
         for i in xnums:
             xkey = "x(%i)-axis" % (i)
             self.x.append(kf.read(sec, xkey))
-            x_name = kf.read(sec, "%s(label)" % (xkey))
+            x_name: int = kf.read(sec, "%s(label)" % (xkey))
```
dormrod: This is marked as `int` but fed straight into a method taking a `str` - is this correct?

src/scm/plams/interfaces/adfsuite/amsanalysis.py
```diff
@@ -61,14 +66,14 @@ def read_data(self, kf, sec):
...
             self.x_names.append(convert_to_unicode(x_name))
             self.x_units.append(convert_to_unicode(kf.read(sec, "%s(units)" % (xkey))))
 
         # Read the y-values
         ykey = "y-axis"
-        y_name = kf.read(sec, "%s(label)" % (ykey))
-        self.y = kf.read(sec, ykey)
+        y_name: str = kf.read(sec, "%s(label)" % (ykey))
```
dormrod: Is there a reason these are not `self.y: Optional[str] = None` in the `__init__`? That would solve a mypy issue

src/scm/plams/interfaces/adfsuite/amsanalysis.py
```diff
@@ -172,10 +177,10 @@ class AMSAnalysisResults(SCMResults):
     _kfext = ".kf"
     _rename_map = {"plot.kf": "$JN" + _kfext}
 
-    def get_molecule(self, *args, **kwargs):
+    def get_molecule(self, *args: Any, **kwargs: Any) -> NoReturn:
         raise PlamsError("AMSAnalysisResults does not support the get_molecule() method.")
 
-    def get_sections(self):
+    def get_sections(self) -> KeysView[Hashable]:
```
dormrod: My IDE implies this is a `str`? That is assumed by `get_all_plots`

src/scm/plams/core/settings.py
```diff
@@ -380,7 +398,7 @@ def get_nested(
 
         return s
 
-    def set_nested(self, key_tuple: Tuple[Hashable, ...], value: Optional[Any], suppress_missing: bool = False):
+    def set_nested(self, key_tuple: Iterable[Hashable], value: Optional[Any], suppress_missing: bool = False) -> None:
```
dormrod: This one needs to be a `Sequence` at least for the indexing, but it calls `contains_nested` which takes a `Tuple[Hashable, ...]` nominally (less restrictive in the code). 

Should it be `Sequence` or changed back?

src/scm/plams/interfaces/adfsuite/amsanalysis.py
ThePauliPrinciple: Not sure how to confirm if it is always a string or not.

src/scm/plams/interfaces/adfsuite/amsanalysis.py
ThePauliPrinciple: I would imagine this is a `str` indeed, considering it's a name.

src/scm/plams/interfaces/adfsuite/amsanalysis.py
ThePauliPrinciple: No particular reason.

src/scm/plams/core/settings.py
ThePauliPrinciple: Likely should be changed back to  `Tuple`.

src/scm/plams/interfaces/adfsuite/ase_calculator.py
```diff
@@ -261,21 +289,21 @@ def ensure_property(self, properties):
             if not property_found:
                 raise NotImplementedError(f"No extractor known for property {prop}")
 
-    def results_from_ams_results(self, ams_results, job_settings):
+    def results_from_ams_results(self, ams_results: "AMSResults", job_settings: Settings) -> None:
         """Populates the self.results dictionary by having extractors act on an AMSResults object."""
         for extractor in self.extractors:
             if extractor.check_settings(job_settings):
                 self.results[extractor.name] = extractor.extract(ams_results, self.atoms)
```
dormrod: Do you know what the behaviour should be here if `self.atoms` is `None`?

src/scm/plams/mol/molecule.py
```diff
@@ -1134,7 +1205,7 @@ def general_supercell(S):
             newmol.translate(sum(i * np.array(vec) for i, vec in zip(index, self.lattice)))
             ret += newmol
 
-        ret.lattice = supercell_lattice
+        ret.lattice = [list(v) for v in supercell_lattice]
```
dormrod, line 1208: (for consistency with an initialised molecule this should be a list of lists)

src/scm/plams/interfaces/adfsuite/ase_calculator.py
ThePauliPrinciple: The behaviour is undefined in that case, obtaining the results is only possible if `calculate` has been called, which will ensure `self.atoms` is set.

dormrod: OK - made it a no-op in that case then

src/scm/plams/interfaces/adfsuite/ase_calculator.py
```diff
@@ -156,33 +165,47 @@ class AMSCalculator(Calculator):
     # counters are a dict as a class variable. This is to support deepcopying/multiple instances with the same name
     _counter: Dict[str, int] = {}
 
-    def __new__(cls, settings=None, name="", amsworker=False, restart=True, molecule=None, extractors=[]):
+    def __new__(  # type: ignore
+        cls: Type[TSelf],
+        settings: Settings,
```
dormrod: This used to be optional but is now required - is this an intentional change?

src/scm/plams/core/basejob.py
```diff
@@ -99,12 +126,12 @@ def __init__(
         if os.path.sep in name:
             raise PlamsError(f"Job name cannot contain {os.path.sep}")
         self._status_log: List[Tuple[datetime.datetime, str]] = []
-        self.status = JobStatus.CREATED
+        self.status: str = JobStatus.CREATED
```
dormrod: Think we should use `JobStatus` class here specifically, to encourage use of the proper enum? The `StrEnum` is just for backwards compatibility

mypy.ini
```diff
@@ -18,7 +20,35 @@ disallow_subclassing_any = False
...
 disallow_any_generics = False
 disallow_untyped_calls = False
-disallow_incomplete_defs = False
-disallow_untyped_defs = False
+# disallow_incomplete_defs = False
+# disallow_untyped_defs = False
 no_implicit_reexport = False
 warn_return_any = False
+
+# Temporary excludes for files with with many tedious errors
```
ThePauliPrinciple: Should this really be put into trunk right now?

src/scm/plams/core/basejob.py
```diff
@@ -454,7 +484,11 @@ class SingleJob(Job):
 
     _filenames = {"inp": "$JN.in", "run": "$JN.run", "out": "$JN.out", "err": "$JN.err"}
 
-    def __init__(self, molecule: Optional[Molecule] = None, **kwargs):
+    def __init__(
+        self,
+        molecule: Optional[Union[Molecule, Dict[str, Molecule], "ChemicalSystem", Dict[str, "ChemicalSystem"]]] = None,
```
dormrod, line 489: This is the type hint for the `AMSJob` class, which passes the `molecule` to here, hence I updated it to match. I don't think there is any issue with doing this

src/scm/plams/core/jobrunner.py
```diff
@@ -325,13 +339,14 @@ def __pbs_running(output: str):  # type: ignore
     config.slurm.commands.getid = __slurm_get_jobid
     config.slurm.commands.running = __slurm_running
 
-    def __init__(self, grid="auto", sleepstep=5, parallel=True, maxjobs=0):
+    def __init__(self, grid: str = "auto", sleepstep: int = 5, parallel: bool = True, maxjobs: int = 0):
```
dormrod, line 342: grid could be a literal of auto, slurm and pbs?

dormrod:
Should we add a `py.typed` file yet? Given the current state of having a partially typed package where the internal mypy issues have not been resolved? 

I think my view would be to add it only when the latter is fixed.

src/scm/plams/core/threading_utils.py
```diff
@@ -127,24 +134,46 @@ def initialized(self) -> bool:
...
+        )
+
+    def run(self) -> None:
         """
         Run thread target in the context copied from the parent thread
         """
         self._context.run(super().run)
+
+
+ContextAwareThread()
```
dormrod: Unintentional diff?

ThePauliPrinciple: Yes

ThePauliPrinciple:
> Should we add a `py.typed` file yet? Given the current state of having a partially typed package where the internal mypy issues have not been resolved?
> 
> I think my view would be to add it only when the latter is fixed.

Yes, let's add that later.

src/scm/plams/core/basejob.py
ThePauliPrinciple: Sure, I'm not actually familiar with typing with enums.

src/scm/plams/core/basejob.py
ThePauliPrinciple: Other than that I do see an issue with molecule having the option of being a chemical system at all, I don't see an issue with this either

src/scm/plams/interfaces/adfsuite/ase_calculator.py
ThePauliPrinciple: It was intentional, but should be reverted as discussed.

dormrod: reverted

src/scm/plams/core/basejob.py
dormrod: done

mypy.ini
dormrod: converted to a to-do

src/scm/plams/core/threading_utils.py
dormrod: removed

dormrod (approved):

PR #303, opened 2025-08-07
Methods for extracting VCD spectra form results

Frankifisu:
I added AMSResults methods for extracting VCD spectral information, mirroring what is already available for IR and Raman.


dormrod (approved):
Ran the example and all looks good 👍

PR #305, opened 2025-08-18
Rosa Bulo (REB) SCMSUITE-- SO107: Fixed small bug in ReactionEquation…

rebadam:
… for charged molecules


examples/balance_equations.py
```diff
@@ -82,6 +82,24 @@ def main():
         coeffs = reaction.balance(min_coeffs)
         print("%s" % (reaction))
 
+    #######################################################################
+    # Test 3
+    # Test equation balancing with charges molecules
```
ThePauliPrinciple: charges molecules -> charged molecules?
Is this a test that is also executed? Considering print statements are used here, I would then expect some file that contains the output that is checked against to also need updating

rebadam: Nee, het is geen test. Het is een example. Dat is waar ik het voorbeeld oorspronkelijk had geplaatst.

rebadam: Hm. I will turn it into a unit test.

dormrod: Thanks - I think the unit test is better and generally preferred. I have just changed it to not run on GitHub (requires installation of CBC solver) - but it will run in our internal infra

dormrod (approved):
Looks good to me. Like the new tests!

PR #315, opened 2025-09-10
PLAMS objects mypy fixes

dormrod:
Enable mypy checking for the PLAMS classes themselves.

Fix remaining issues which are not in the specific module ignores, either by small code changes or adding ignore statements.

Replace all old bare-ignore statements with specific exclusions.

### DIFF
--- a/src/scm/plams/interfaces/adfsuite/ams.py
+++ b/src/scm/plams/interfaces/adfsuite/ams.py
@@ -845,7 +845,7 @@ class AMSResults(Results):
 
         x = np.concatenate(x).ravel()
 
-        return x, y, labels  # type: ignore[]
+        return x, y, labels  # type: ignore[return-value]
 
     def get_phonons_thermodynamic_properties(
         self, temperature_unit: str = "K", properties_unit: List[str] = ["hartree", "kB"]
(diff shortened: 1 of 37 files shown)


src/scm/plams/interfaces/adfsuite/amsanalysis.py
```diff
@@ -300,7 +300,7 @@ def _serialize_mol(self) -> None:
         """
         from scm.plams import AMSJob
 
-        systems = AMSJob._serialize_molecule(self)
+        systems = AMSJob._serialize_molecule(self)  # type: ignore[arg-type]
```
dormrod, line 303: This for example is a bad pattern imo... reusing a method of `AMSJob` in `SCMJob` but seems quite common and not worth a refactor here

src/scm/plams/interfaces/adfsuite/ams.py
```diff
@@ -272,6 +272,8 @@ def get_molecule(self, section: str, file: str = "ams") -> Molecule:
         sectiondict = self.read_rkf_section(section, file)
         if sectiondict:
             return Molecule._mol_from_rkf_section(sectiondict)
+        else:
+            raise ValueError(f"Could not retrieve molecule from rkf section '{section}' for file '{file}.rkf'")
```
dormrod, line 276: This is technically a small breaking change as this previously would have returned `None`, but I think it is better to raise the error

src/scm/plams/interfaces/adfsuite/amsanalysis.py
ThePauliPrinciple: Is it fixed if the `self` argument of `AMSJob._serialize_molecule` is annotated to be `Union[AMSJob, SCMJob]`?

dormrod: I don't think I want to encourage this behaviour - I think it should just be refactored ultimately

PR #316, opened 2025-09-11
Add typed versions of KFFile.read 

ThePauliPrinciple:
This PR adds `KKFFile.read_{int,ints,real,reals,logical,logicals,string}` which do the exact same thing as `KFFile.read` except throws an error if the variable read was not of the right type. The plural versions always return a list, even if there is a single value. (`KFFile.read(..., return_as_list=True)`).

### DIFF
--- a/src/scm/plams/tools/kftools.py
+++ b/src/scm/plams/tools/kftools.py
@@ -21,7 +21,7 @@ from typing import (
 )
 
 import numpy as np
-from scm.plams.core.errors import FileError
+from scm.plams.core.errors import FileError, PlamsError
 from scm.plams.core.functions import log
 from scm.plams.core.private import saferun
 import numpy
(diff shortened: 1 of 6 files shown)


src/scm/plams/tools/kftools.py
```diff
@@ -373,6 +398,97 @@ def read(self, section: str, variable: str, return_as_list: bool = False) -> TRe
...
+    def read_ints(self, section: str, variable: str) -> List[int]:
+        """Obtain a list of integers for a *variable* located in *section*.
+
+        :param section: Section where the variable is located
+        :param variable: Variable to obtain
+        :raises KFTypedReadError: Error is raised if the *variable* in *section* does not contain a list of integers
+        :return: The list of integers of *variable* located in *section*
+        """
+        ret = self.read(section, variable, return_as_list=True)
+        if not isinstance(ret, List) or len(ret) == 0 or not isinstance(ret[0], int):
```
dormrod: Just checking we want to raise an error if this is an empty list? There's no situation where some supposed integer list on the KF could be empty?

src/scm/plams/tools/kftools.py
```diff
@@ -373,6 +398,97 @@ def read(self, section: str, variable: str, return_as_list: bool = False) -> TRe
...
+        :param variable: Variable to obtain
+        :raises KFTypedReadError: Error is raised if the *variable* in *section* does not contain a list of logicals/bools
+        :return: The list of logicals/bools of *variable* located in *section*
+        """
+        ret = self.read(section, variable, return_as_list=True)
+        if not isinstance(ret, List) or len(ret) == 0 or not isinstance(ret[0], bool):
+            raise KFTypedReadError(section, variable, ret, "List[bool]")
+        return ret
+
+    def read_string(self, section: str, variable: str) -> str:
```
dormrod, line 479: Any benefit to having a `read_strings` method which takes a delimiter to split on? e.g. `\x00`

src/scm/plams/tools/kftools.py
```diff
@@ -373,6 +398,97 @@ def read(self, section: str, variable: str, return_as_list: bool = False) -> TRe
             ret = [ret]
         return ret
 
+    def read_int(self, section: str, variable: str) -> int:
```
dormrod, line 401: I know these are quite straightforward, but might as well add a small unit test case to https://github.com/SCM-NV/PLAMS/blob/trunk/unit_tests/test_kftools.py#L13 ?

src/scm/plams/tools/kftools.py
ThePauliPrinciple: Yes, but the libbase version does some complicated stuff (some enum, also with options to split based on number of chars etc) and didn't want to end up with a version that works differently. writing `kf.read_string().split('\x00')` is also not that bad.

src/scm/plams/tools/kftools.py
ThePauliPrinciple: You are allowed to have an empty list, I will add an explicit read for the type in case of an empty list.

src/scm/plams/tools/kftools.py
ThePauliPrinciple: I added some

dormrod: Nice

src/scm/plams/tools/kftools.py
dormrod: OK agreed, leave as-is

dormrod (approved):

PR #318, opened 2025-09-12
How to use uv for plams development

GiulioIlBen:
Added a description for uv usage in the CONTRIBUTING.md

Added skip_if_no_scm_libbase to 2 tests to avoid installation error.


CONTRIBUTING.md
```diff
@@ -45,4 +45,35 @@ These include:
...
+PYTHONPATH=${PYTHONPATH}:./src:./unit_tests uv run pytest unit_tests/
+```
+To add optional dependencies use (default dependencies are difficult to be added):
+```bash
+uv add pigeon-jupyter --optional ml
+```
+
+### Fix the code before submitting it
+
+Check the file /home/benedini/PLAMS/.github/workflows/ci.yml,
```
dormrod: Update to be the link on GitHub, and use a proper hyperlink

dormrod (commented):
Do we want to include the lock file in the repo? To me this implies that the user should use all the package deps that you have specified here. Whilst this makes it more reproducible, it is also less flexible? Particularly as these will have been chosen because of your python version (3.9?)

For example, it would be interesting to know if a more recent version of X package actually causes an issue, as then a fix can be made / the dependency range in the toml file updated.

Devs could also just run `uv lock` in their local cloned repo as part of the development flow?

CONTRIBUTING.md
GiulioIlBen: Thank you

GiulioIlBen:
True.

As a solution maybe the lock file should be added in .gitignore?

I am not sure how would be the ideal workflow for plams development, but still I would improve the exsiting CONTRIBUTING.md

dormrod:
> True.
> 
> As a solution maybe the lock file should be added in .gitignore?
> 
> I am not sure how would be the ideal workflow for plams development, but still I would improve the exsiting CONTRIBUTING.md

Yes you could add to the .gitignore. I think basically the same as what you have, but first add a step where you run `uv lock`?

dormrod:
> > True.
> > As a solution maybe the lock file should be added in .gitignore?
> > I am not sure how would be the ideal workflow for plams development, but still I would improve the exsiting CONTRIBUTING.md
> 
> Yes you could add to the .gitignore. I think basically the same as what you have, but first add a step where you run `uv lock`?

Sync does an implicit lock.

Agreed to add to .gitignore for now, but as pointed out see this thread for context and options for `uv sync` when including the lock file, to test different dependency ranges: https://github.com/astral-sh/uv/issues/10730

dormrod (approved):

PR #319, opened 2025-09-12
Added conformers interface to plams

GiulioIlBen:
The plams interface instead of living in `scm.conformers` has moved to `scm.plams`. This is in accordance with what @mhellstr and I discussed, we think is a cleaner approach and more inline with plams design.

Added method `plot_conformers` to `ConformersResults` .

Updated the notebooks examples but not the .py files yet. 

Let me know what do you think.


examples/COSMORSConformers/cosmors_conformers.ipynb
```diff
@@ -18,14 +18,14 @@
...
-      "PLAMS working folder: /path/plams/examples/COSMORSConformers/plams_workdir\n"
+      "PLAMS working folder: /home/benedini/PLAMS/examples/COSMORSConformers/plams_workdir.002\n"
      ]
     }
    ],
    "source": [
     "from scm.plams.recipes.adfcosmorsconformers import ADFCOSMORSConfJob, ADFCOSMORSConfFilter\n",
     "from scm.plams import Molecule, from_smiles, Settings, init, config, JobRunner\n",
-    "from scm.conformers import ConformersJob\n",
+    "from scm.plams import ConformersJob\n",
```
dormrod, line 28: You need to run `generate_example.sh` on all these examples to get these updates into the docs

src/scm/plams/interfaces/adfsuite/conformers.py
```diff
@@ -0,0 +1,280 @@
...
+    read_all_molecules_in_xyz_file,
+    requires_optional_package,
+)
+from scm.plams.core.settings import Settings
+from scm.plams.mol.molecule import Molecule
+from scm.plams.tools.kftools import KFFile
+from scm.plams.tools.plot import plot_molecule
+from scm.plams.tools.units import Units
+
+from .ams import AMSJob, AMSResults
```
dormrod: (convert to absolute import)

src/scm/plams/interfaces/adfsuite/conformers.py
```diff
@@ -0,0 +1,280 @@
...
+            raise ValueError(f"{path=} not valid")
+
+        assert path.exists()
+        molecules = read_all_molecules_in_xyz_file(path)
+        for m in molecules:
+            m.properties = self.parse_comment(m.properties.pop("comment", ""))
+        self.molecules = list(sorted(molecules, key=lambda m: m.properties["Energy in kcal/mol"], reverse=False))
+
+    @property
+    def energies(self):
```
dormrod: Please could you add any missing return type hints (including `-> None`) for all these methods whilst we are doing this refactor?

src/scm/plams/interfaces/adfsuite/conformers.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def __init__(self, name="conformers", molecule=None, **kwargs):
+        Job.__init__(self, name=name, **kwargs)
+
+        if molecule is None:
+            self.molecule = None
+        elif isinstance(molecule, Molecule):
+            self.molecule = molecule.copy()
+        else:
+            raise NotImplementedError("TODO: Implement multiple molecules input.")
```
dormrod: Please update this to something more like - `ValueError(f"If specified, molecule but be a single Molecule instance, but was '{molecule}'"`

Also add a comment about UCS... this probably needs to support it in future

doc/source/examples/COSMORSConformers/cosmors_conformers.ipynb.rst
```diff
@@ -8,14 +8,14 @@ Initial imports
...
-   from scm.conformers import ConformersJob
+   from scm.plams import ConformersJob
 
    # this line is not required in AMS2025+
    init();
 
 ::
 
-   PLAMS working folder: /path/plams/examples/COSMORSConformers/plams_workdir
+   PLAMS working folder: /home/benedini/PLAMS/examples/COSMORSConformers/plams_workdir.002
```
dormrod: The generate_example should have sanitised these strings - I will have to look why this did not work

examples/ConformersGeneration/ConformersGeneration.py
```diff
@@ -1,55 +0,0 @@
-#!/usr/bin/env amspython
-from scm.plams import *
-from scm.conformers import ConformersJob, ConformersResults
```
dormrod, line 3: This is used as an example in the nightly tests - so either that needs to be updated or this can be added back in

src/scm/plams/interfaces/adfsuite/conformers.py
```diff
@@ -0,0 +1,315 @@
...
+    def get_energy_landscape(self) -> AMSResults.EnergyLandscape:
+        # TODO how is this possible?
+        el = AMSResults.EnergyLandscape(None)
+        for energy, mol in zip(self.get_energies(), self.get_conformers()):
+            state = AMSResults.EnergyLandscape.State(el, None, energy, mol, 1, False)
+            el._states.append(state)
+        return el
+
+    @requires_optional_package("matplotlib")
+    def plot_conformers(
```
dormrod, line 187: Should this be here or in tools/plot for consistency?

CONTRIBUTING.md
```diff
@@ -22,6 +22,7 @@ top of things.
 * Make commits of logical units
 * Make sure your commit messages are informative
 * Make sure you have added the necessary tests for your changes
+* Add examples: take a look at the corresponding [README.md](./examples/README.md)
```
dormrod, line 25: Also update CHANGELOG (here and for your change!)

doc/source/examples/COSMORSConformers/cosmors_conformers.ipynb.rst
dormrod: Fixed

src/scm/plams/__init__.py
```diff
@@ -52,14 +52,18 @@
...
+from scm.plams.interfaces.adfsuite.conformers import ConformersJob, ConformersResults
 from scm.plams.interfaces.adfsuite.crs import CRSJob, CRSResults
 from scm.plams.interfaces.adfsuite.densf import DensfJob, DensfResults
 from scm.plams.interfaces.adfsuite.fcf import FCFJob, FCFResults
 from scm.plams.interfaces.adfsuite.forcefieldparams import (
     ForceFieldPatch,
     forcefield_params_from_kf,
 )
-from scm.plams.interfaces.adfsuite.inputparser import get_system_blocks_as_molecules_from_input, input_to_settings
+from scm.plams.interfaces.adfsuite.inputparser import (
```
dormrod, line 63: For future PRs, please can you turn off the auto-reordering of imports? It bloats the diff and makes it a bit harder to review

src/scm/plams/interfaces/adfsuite/conformers.py
```diff
@@ -0,0 +1,315 @@
...
+from scm.plams.interfaces.adfsuite.ams import AMSJob, AMSResults
+from scm.plams.mol.molecule import Molecule
+from scm.plams.tools.kftools import KFFile
+from scm.plams.tools.plot import plot_molecule
+from scm.plams.tools.units import Units
+
+__all__ = ["ConformersJob", "ConformersResults"]
+
+
+class ConformersCollection:
```
dormrod, line 22: I see that this is actually not just a straight "lift and shift".

You have replaced the old `Conformers` usage to remove the dependency on `scm.conformers` with this class. This seems like it might have consequences - will some functionality be lost? The plan is I assume to remove the old implementation in `scm.conformers`?

Might be worth asking @rebadam if you have not already discussed with her.

src/scm/plams/interfaces/adfsuite/conformers.py
```diff
@@ -0,0 +1,315 @@
...
+    "in memory conformers collection"
+
+    def __init__(self, path_in: str):
+        path = Path(path_in)
+        if path.is_dir():
+            path = path / "conformers.xyz"
+        if not path.name.endswith("conformers.xyz"):
+            raise ValueError(f"{path=} not valid")
+
+        assert path.exists()
```
dormrod, line 32: (don't use assert in production code)

src/scm/plams/interfaces/adfsuite/conformers.py
```diff
@@ -0,0 +1,315 @@
...
+        molecules = read_all_molecules_in_xyz_file(str(path))
```
dormrod, line 33: Using the xyz files for example loses any bonding information?

src/scm/plams/interfaces/adfsuite/conformers.py
rebadam: I am quite ok with the PLAMS wrapper around the conformers tool being very basic. Though I would prefer if we could point the user to the conformers library we already have if they want more functionality. But I don't like so much that you wrote a whole new Conformers class. It is very basic, which is good, but I am afraid people will start adding to it, which will result in a lot of double coding. I would prefer if the functionality of the ConformersContainer class would just be folded into the Results class directly. The `conformers` property of the results class could just be deprecated.

src/scm/plams/interfaces/adfsuite/conformers.py
rebadam: Yes, I agree here. It would be super simple to just read from the conformers.rkf file, which also has higher precision for the coordinates. The RKF file just has the geometries and energies in the History section.

CONTRIBUTING.md
GiulioIlBen: should I mention that I changed the CONTRIBUTING.md file in the CHANGELOG?

GiulioIlBen:
If I run ConformersJob from a different Python environment, which Python environment does the workflow use—the default amspython or the one I launched it from?

dormrod:
> If I run ConformersJob from a different Python environment, which Python environment does the workflow use—the default amspython or the one I launched it from?

It will be called with:
```
"$AMSBIN"/amspython -c "from scm.utils.workflows import run_workflow; from scm.conformers import AMSConformers; run_workflow(AMSConformers)" $@
```
so amspython

PR #320, opened 2025-09-12
Giulio benedini/log discord

GiulioIlBen:
I find it very useful to log when a long workflow is finished and send a message via discord.

I have implemented a simple function: `log_message_via_discord`.

Check it out if you could find it useful.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,10 +6,10 @@ readme = "README.md"
 requires-python = ">=3.8"
 license = { text = "LGPLv3" }
 authors = [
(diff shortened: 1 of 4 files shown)


dormrod:
Closing as this change does not really sit within PLAMS.

However, it is a great idea. I think instead though, in PLAMS we could add a callback into the job status change, so that the user can supply a function which performs an action on success/failure. 

This kind of feature would be nice to also have in the AMS GUI, so also when running a notification can be sent (via a variety of channels). So I will close the PR for now, and add it to that project

PR #322, opened 2025-09-16
Some typing fixes

ThePauliPrinciple:
- type ignore Atom.things - Some algorithms add temporary attributes to the Atom class, this can't be typed as is
- add casts for readrkf/readkf - Some `cast(Type, variable)` added to add type hints to structured data read from IO
- add `Job.get_path` - convenient compared to `Job.path` attribute that can be `None` in some cases.
- add specific subclasses to results/job attributes - `Job`/`Results` pairs need their `results` and `job` attributes typed to their pair so that it can be known what methods/attributes are available

### DIFF
--- a/src/scm/plams/core/results.py
+++ b/src/scm/plams/core/results.py
@@ -214,7 +214,7 @@ class Results(ApplyRestrict):
             old = old.replace("$JN", self.job.name)
             new = new.replace("$JN", self.job.name)
             if old in self.files:
-                os.rename(opj(self.job.path, old), opj(self.job.path, new))
+                os.rename(self.job.get_path() / old, self.job.get_path() / new)
                 self.files[self.files.index(old)] = new
         self.refresh()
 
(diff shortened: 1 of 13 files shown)


src/scm/plams/core/results.py
```diff
@@ -321,10 +321,10 @@ def rename(self, old: str, new: str) -> None:
         new = new.replace("$JN", self.job.name)
         self.refresh()
         if old in self.files:
-            os.rename(opj(self.job.path, old), opj(self.job.path, new))
+            os.rename(self.job.get_path() / old, self.job.get_path() / new)
             self.files[self.files.index(old)] = new
         else:
-            raise FileError(f"File {old} not present in {self.job.path}")
+            raise FileError(f"File {old} not present in {str(self.job.get_path())}")
```
dormrod, line 327: `str` unnecessary here (and elsewhere) I think, default for f-string

dormrod:
Looks reasonable, probably just need to patch the unit test mocks here: https://github.com/SCM-NV/PLAMS/blob/trunk/unit_tests/test_amsresults.py#L24

dormrod (approved):

PR #324, opened 2025-09-17
Drop Unified prefix from UCS and related classes in scm.libbase

robertrueger:
The big rename on amshome is coming up:
https://git.scm.com/SCM/amshome/pulls/107

This is the PLAMS part of it. I guess this needs to happen in PLAMS first, so I can use the merge commit that closes this PR to update the submodule on amshome when merging the PR there?


dormrod (approved):
Looks good to me. And yes I think this needs to go in first, the the submodule in amshome can be updated in your PR there

PR #333, opened 2025-10-14
Rosa Bulo (REB) SCMSUITE-- SO107: SCM lattice added to SDF and XYZ fo…

rebadam:
…rmats

- XYZTrajectoryFile and XYZHistoryFile can now read and write the SDF format lattice
- RKFTrajectoryFile and RKFHistoryFile now use scmlib.KFFile as default
- All trajectories no longer require filename as argument in the __init__, so that it is easier to supply a non-default file_object.
- The binary file objects now work if the mode specified by user does not contain 'b'.


unit_tests/test_trajectories.py
```diff
@@ -32,3 +37,44 @@ def test_frames(self, conformers_rkf):
             # But the labels still match considering bond connectivity
             assert not np.allclose(crds, input_mol.as_array())
             assert mol.label(3) == input_mol.label(3)
+
+
+class TestXYZFiles:
+
+    # def test_historyfile(self, trajectory_xyz):
```
nfaguirrec: I wrote this test, but it is failing in getting the length of the history file. Could you please fix it and modify this test to make it equivalent to test_trajectoryfile?

The error I get is:
```
    def test_historyfile(self, trajectory_xyz):
    
        xyz = XYZHistoryFile(trajectory_xyz)
        mol = xyz.get_plamsmol()
    
        xyzout = XYZHistoryFile('new.xyz',mode='w')
>       for i in range(xyz.get_length()) :

test_trajectories.py:50: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../src/scm/plams/trajectories/trajectoryfile.py:378: in get_length
    crd, cell = self.read_next(read=False)
../src/scm/plams/trajectories/xyzfile.py:164: in read_next
    return self._move_cursor_without_reading()
../src/scm/plams/trajectories/trajectoryfile.py:217: in _move_cursor_without_reading
    if self._is_endoffile():
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <scm.plams.trajectories.xyzhistoryfile.XYZHistoryFile object at 0x7939a489a1f0>

    def _is_endoffile(self):
        """
        If the end of file is reached, return coords and cell as None
        """
        end = False
        line = self.file_object.readline()
        if len(line) == 0:
            end = True
            return end
>       nats = int(line.split()[0])
E       IndexError: list index out of range

../src/scm/plams/trajectories/xyzhistoryfile.py:108: IndexError
```

unit_tests/test_trajectories.py
```diff
@@ -32,3 +37,44 @@ def test_frames(self, conformers_rkf):
...
+        for i in range(trajectory_file.get_length()) :
+            crd,cell = trajectory_file.read_frame(i,molecule=mol)
+            xyzout.write_next(molecule=mol)
+        xyzout.close()
+
+        with open('new.xyz') as f: output = f.read()
+        expectedOutput = """\
+3
+
+       O        -0.1995727721        -0.2577814972        -0.0000000000
```
nfaguirrec: This test works, but there is an annoying space at the end of the coordinates lines that should be nice to remove. I added the space here manually for the test to pass. The space is added somewhere where plams save the xyz info to strings.

nfaguirrec:
I added a test but there are some things to fix. Please, see my comments in the code

rebadam:
I changed the test, so that it now:
- No longer writes to file. For that I had to change the classes a bit, so that they accept StringIO objects
- Also tests SDFTrajectoryFile
- Uses the conformers.rkf, so that it is not necessary to add a new file to the database. The code reads te RKF file, adds lattice vectors, writes an XYZ trajectory with VEC entries, and reads that XYZ file again.

nfaguirrec (approved):
Now works fine! Nice!

PR #339, opened 2025-11-20
Rosa Bulo (REB) SCMSUITE-- SO107: Fixed bug in Molecule.reorder

rebadam:
- The reorder method failed for very simple cases, because it only compares edges (bonds) and not the nodes. The graph passed to NetworkX now contains edges between each node and itself, containing identifying data for the node.
- I also changed the code so that in creating unique identifiers for the atoms, which are used to map the graphs, the depth of the dfs can be adjusted. The argument was already there, but it was unused, which I expect was a bug.


dormrod (approved):
Good catch, lgtm!

PR #343, opened 2025-12-12
add nmr example SO--

mhellstr:


mhellstr:
@dormrod the way to do CPL and NMR is quite ugly. Even if one were to create "CPLJob" and "NMRJob", it wouldn't behave as expected, because you can't get the NMR results from the NMRJob (instead it would modify a previously run job).

So my approach was to just dump it into the postamble_lines.

I also realized I have no idea what AMSspectra actually does to add the couplings, so I just deferred to the GUI tutorial for that.

dormrod:
> @dormrod the way to do CPL and NMR is quite ugly. Even if one were to create "CPLJob" and "NMRJob", it wouldn't behave as expected, because you can't get the NMR results from the NMRJob (instead it would modify a previously run job).
> 
> So my approach was to just dump it into the postamble_lines.
> 
> I also realized I have no idea what AMSspectra actually does to add the couplings, so I just deferred to the GUI tutorial for that.

I see what you mean. Fair enough. I made tiny changes to the notebook but otherwise fine.

dormrod (approved):

PR #344, opened 2025-12-16
Use plams Units instead of libbase Units for some conversions.

mirkofranchini:


dormrod (approved):
Looks good, I just also updated the tutorials in the documentation.

In general after changing a Jupyter notebook example, just run `examples/generate_example.sh` and this will generate the .py file and also update the tutorial for free

PR #345, opened 2025-12-17
Rosa Bulo (REB) SCMSUITE-- SO107: Fixed bug RKFHistoryFile nob-block …

rebadam:
…property reading

- Fixed a bug where properties that were not stored as block items (because they were stored as lists) would be interpreted as block items if the first encountered list had length 1. This bug was present in both RKFTrajectoryFile and RKFHistoryFile.
- Added a test for this property reading.


rebadam:
There is a windos test that fails, but it does not appear to have anything to do with the changes I made. Is it because of the temporary directory I created in my test?

rebadam:
I made a small change (just added a print statement), and now the tests pass, so I guess it was a random error.

dormrod (approved):

PR #353, opened 2026-01-21
Rosa Bulo (REB) SCMSUITE-- SO107: input_to_dict now recursive over su…

rebadam:
…bengines

- input_to_dict of AMS input now recursively converts subengines to dictionaries This behavior is different from libbase.InputFile.to_dict(), but the behavior was already different before.
- The reason for this change is a future change I am making in the libbase.InputFile, which keeps the Engine block unchanged upon first processing of the input text. This messes up a PLAMS unittest, which creates an AMSJob from input, and then recreates the input. This job counted on the engine text lines being stripped, which will no longer be the case with the coming changes. Adding new spaces at the start of non-stripped lines makes no sense, so as a work around I made sure the engine block is no longer read as lines, but as a dictionary.

### DIFF
--- a/src/scm/plams/interfaces/adfsuite/inputparser.py
+++ b/src/scm/plams/interfaces/adfsuite/inputparser.py
@@ -178,32 +178,57 @@ def input_to_settings(
     :param parser: use specific parser or defaults to internal parser if ``None``
     :return: |Settings| object with the structure of the parsed AMS input
     """
+
+    def _separate_engine_settings(lines: List[str], depth: int = 0) -> Tuple[Settings, List[str]]:
+        """
+        Recursively resolve the engine settings and add them to the rest
+        """
+        input_settings = Settings()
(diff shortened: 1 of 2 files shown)


rebadam:
Hi David. As stated above I made the input_to_dict function recursive. I ran quite a lot of tests, so I think it the change is fine. But I did not discuss it with anyone, so can you think of any reason why this may be undesirable, or of something it might break after all?

dormrod:
> Hi David. As stated above I made the input_to_dict function recursive. I ran quite a lot of tests, so I think it the change is fine. But I did not discuss it with anyone, so can you think of any reason why this may be undesirable, or of something it might break after all?

Thanks Rosa, I also tested locally and it seems fine to me. I can't think of a reason why this would cause issues going forwards

dormrod (approved):

PR #376, opened 2026-03-12
AMSWorker Support for ChemicalSystem

dormrod:
# Description

This PR adds support for passing `scm.base.ChemicalSystem` to `AMSWorker`, alongside PLAMS Molecule, and adds `AMSWorkerResults.get_input_system()` / `get_main_system()` in line with `AMSResults`.

# Changes

Main updates:

* Added `ChemicalSystem` handling in `AMSWorker._prepare_system()` via type branching (Molecule vs ChemicalSystem), including serialization of coordinates, charge, lattice, bonds, and atomicInfo
* Added lazy initialization in `AMSWorkerResults` for `_input_molecule` and `_input_system`, with conversion between representations using same logic as in scm.utils (not used to avoid dependency)
* Added `AMSWorkerResults.get_input_system()` and `AMSWorkerResults.get_main_system()`
* Existing `AMSWorkerResults.get_input_molecule` and `get_main_molecule` now always return copies
* Added unit tests covering the new ChemicalSystem pathways and parity with existing Molecule behaviour.
* Integration test to be added in AMSHOME


src/scm/plams/interfaces/adfsuite/amsworker.py
```diff
@@ -983,34 +1030,83 @@ def _solve(
...
+        angstrom_to_bohr = Units.conversion_ratio("Angstrom", "Bohr")
+
+        if _has_scm_chemsys and isinstance(molecule, ChemicalSystem):
+            from scm.base import AtomAttributes
+
+            symbols = np.asarray([atom.symbol for atom in molecule.atoms])
+            coords = np.asarray(molecule.coords)
+            charge = molecule.charge
+
+            atomicInfo = []
```
robertrueger: Check the ``cs.atom_EOL_string()`` method. I builds exactly this string of regions + atomic properties that you need here.

dormrod: Great thanks, that works perfectly and is much simpler

robertrueger (approved):
Looks good to me now. When this is on amshome trunk, I'll update some of the regression tests to also exercise the new code paths.

PR #386, opened 2026-03-25
Fix test_plot_molecule

GiulioIlBen:
Explicitly enable axes in plot_molecule() when keep_axis=True. Since ase.plot_atoms novel version  seems to modify the default value.

### DIFF
--- a/src/scm/plams/tools/plot.py
+++ b/src/scm/plams/tools/plot.py
@@ -335,7 +335,6 @@ def plot_molecule(
         ax.axis("on")
     else:
         ax.axis("off")
-        
 
     return ax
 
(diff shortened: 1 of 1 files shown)


src/scm/plams/tools/plot.py
```diff
@@ -333,6 +333,8 @@ def plot_molecule(
 
     if not keep_axis:
         ax.axis("off")
+    else:
```
mhellstr: A general comment is that if there is no particular reason to deviate, it is easier to read
```
if something:
...
else:
...
```

rather than
```
if not something:
...
else:
...
```
to avoid double-negatives.

dormrod (approved):

Issue #398, opened 2026-05-13
Idea: introducing generic typing in `MultiJob` for `self.children`


SiebeLeDe:
# Main Proposal

My proposal is to include generic typing in the [`MultiJob`](https://github.com/SCM-NV/PLAMS/blob/trunk/src/scm/plams/core/basejob.py) class to allow for correct typing behavior in `self.children` overwrites.

# Current issue

Say I declare my own class, `TetramerEDAJob`, that inherets from the `MultiJob` and I want to specify which kind of jobs it will contain. Doing so, gives me typing information in my code editor about the available methods which prevents me from making stupid typos and working with the wrong return type (e.g., float instead of np.array containing floats (I use VS code with Python extensions enabled). In this case, I have a MultiJob that contains `ADFFragmentJob` (also inhereting from a MultiJob).

```python
class TetramerEDAJob(MultiJob):
    def __init__(self, fragments: List[ChemicalSystem], complex: ChemicalSystem, settings: Settings, print_detailed_output: bool = True, **kwargs) -> None:
        super().__init__(settings=settings, **kwargs)
        self.children: List[ADFFragmentJob] = [] 
```

relying on 

```python
class MultiJob(Job):
    def __init__(self, children: Optional[List[Job]] = None, childrunner: Optional["JobRunner"] = None, **kwargs: Any):
        Job.__init__(self, **kwargs)
        self.children: List[Job] = [] if children is None else children
```

So, I have a `MultiJob` consisting of `MultiJob`s. However, Pylance raises an "override" error saying that the type (`List[ADFFragmentJob]`) is not invariant since it expects an instance of the `Job` class specifically:

```
"children" overrides symbol of same name in class "MultiJob"
  Variable is mutable so its type is invariant
    Override type "List[AMSJob | ADFFragmentJob]" is not the same as base type "List[Job]")
```

But, the idea should be that the MultiJob may contain any kind of a (user-defined) Job that inherets from `Job`; it should not be limited to`Job` specifically (which is actually an abstract base class and should never be instanced, i.e., `job_instance = Job(name=...)`).

[...]

PR #399, opened 2026-05-28
Introducing generic typing in MultiJob for self.children

SiebeLeDe:
Solves #398 about type errors because the self.children attribute in `MultiJob` was not generically typed.

# Changes

- Added Generic Job type to the `MultiJob` class, including the methods `new_children`, `other_jobs`, `remove_child`, and dunder method `__iter__`
- Changed the `self.children` type from `List[J]` to `Union[List[J], Dict[J]]`

No (new) tests have been written since these changes only affect the Python linter and typechecker in the text editor.

### DIFF
--- a/src/scm/plams/core/basejob.py
+++ b/src/scm/plams/core/basejob.py
@@ -836,9 +836,9 @@ class MultiJob(Job, Generic[J]):
     Private attributes ``_active_children`` and ``_lock`` are essential for proper parallel execution. Please do not modify them.
     """
 
-    def __init__(self, children: Optional[List[J]] = None, childrunner: Optional["JobRunner"] = None, **kwargs: Any):
+    def __init__(self, children: Optional[Union[List[J], Dict[str, J]]] = None, childrunner: Optional["JobRunner"] = None, **kwargs: Any):
         Job.__init__(self, **kwargs)
-        self.children: List[J] = [] if children is None else children
+        self.children: Union[List[J], Dict[str, J]] = [] if children is None else children
         self.childrunner = childrunner
         self._active_children = 0
         self._lock = threading.Lock()
--- a/src/scm/plams/core/basejob.py
+++ b/src/scm/plams/core/basejob.py
@@ -21,6 +21,8 @@ from typing import (
     Any,
     Iterator,
     Set,
+    Generic,
+    cast,
 )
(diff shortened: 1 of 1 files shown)


src/scm/plams/core/basejob.py
```diff
@@ -955,14 +957,14 @@ def delete(self) -> None:
         if self.status != JobStatus.CREATED:
             self.results.wait()
 
-        for child in [c for c in self.children]:
+        for child in [c for c in self.children] if isinstance(self.children, list) else self.children.values():
```
dormrod: It looks like there is currently no unit-test coverage for MultiJobs with a dictionary of child jobs. 

This would cause an error with e.g. a modified version of https://github.com/SCM-NV/PLAMS/blob/trunk/unit_tests/test_basejob.py#L1200 with:
```
RuntimeError: dictionary changed size during iteration
```

This is an existing bug which predates your change, but it would be nice if you could fix this line with `list(self.children.values())` or similar.

I can also later add the missing unit test coverage, by subclassing https://github.com/SCM-NV/PLAMS/blob/trunk/unit_tests/test_basejob.py#L915

dormrod (changes requested):
Thanks very much for submitting this change. 

Please also just update the [Changelog](https://github.com/SCM-NV/PLAMS/blob/trunk/CHANGELOG.md) with a summary of this change, and the bug fix requested

SiebeLeDe:
I see my note to the changelog was not included in the changelog. I had written in the markdown file: 

* `MultiJob` now supports generic `Job` types for the `self.children` attribute

I also see that my ruff formatter reformatted the basejob.py file, which was not my intention. My apologies. Is there something I can do to fix this (i.e. redo the formatting and add the change to the changelog)?

dormrod:
> I see my note to the changelog was not included in the changelog. I had written in the markdown file:
> 
> * `MultiJob` now supports generic `Job` types for the `self.children` attribute
> 
> I also see that my ruff formatter reformatted the basejob.py file, which was not my intention. My apologies. Is there something I can do to fix this (i.e. redo the formatting and add the change to the changelog)?

No worries, I already ran the black formatter on the code and will add the small change to the changelog

SiebeLeDe:
Thank you!

PR #404, opened 2026-06-12
add timestep from OUTCAR file from VASP

Fra-me:
Added timestep reading from OUTCAR file from VASP, together with unit tests.


src/scm/plams/tools/converters.py
```diff
@@ -7,8 +7,8 @@
 from scm.plams.mol.molecule import Molecule
 from scm.plams.tools.kftools import KFFile
 from scm.plams.tools.units import Units
-from scm.plams.core.functions import requires_optional_package
-from scm.plams.trajectories.rkffile import RKFTrajectoryFile
+from scm.plams.core.functions import requires_optional_package, log
+from scm.plams.trajectories.rkffile import RKFTrajectoryFile
```
dormrod: For the formatting, install black:
```
amspython -m pip install "black[jupyter]"==24.8.0
```
then run:
```
amspython -m black -t py38 -l 120 .  
```
in the PLAMS root.

This will automatically format all the code in the accepted way

src/scm/plams/tools/converters.py
```diff
@@ -270,6 +295,21 @@ def vasp_output_to_ams(
...
+    # IBRION = 0 means molecular dynamics, which determines both the timestep and the task.
+    if timestep is None or task is None:
+        ibrion, potim = _read_md_params_from_outcar(outcar, max_lines=100)
+        is_md = ibrion == 0
+        if timestep is None:
+            if is_md:
+                timestep = potim
+            else:
+                timestep = 0.25  # not MD: Time is still written, but the value is irrelevant
+                log(f"{outcar} is not a molecular-dynamics run (IBRION={ibrion}); "
```
dormrod: Pass level 5 to this log call to reduce the verbosity

src/scm/plams/tools/converters.py
```diff
@@ -219,13 +219,34 @@ def _postprocess_vasp_amsrkf(kffile: str, outcar: str) -> None:
...
+
+
 def vasp_output_to_ams(
     vasp_folder: str,
     wdir: Optional[str] = None,
     overwrite: bool = False,
     write_engine_rkf: bool = True,
     task: Optional[str] = None,
-    timestep: float = 0.25,
+    timestep: Optional[float] = None,
```
dormrod, line 252: Update the changelog with details of the changes in this PR: https://github.com/SCM-NV/PLAMS/blob/trunk/CHANGELOG.md

src/scm/plams/tools/converters.py
```diff
@@ -219,13 +219,34 @@ def _postprocess_vasp_amsrkf(kffile: str, outcar: str) -> None:
         kf.save()
 
 
+def _read_md_params_from_outcar(outcar_path, max_lines=100):
```
dormrod: Also add type hints here, as good practice

unit_tests/test_converters.py
```diff
@@ -0,0 +1,53 @@
...
+    return vasp_folder / "NpT"
+
+@pytest.fixture
+def relaxation_vasp(vasp_folder):
+    return vasp_folder / "relaxation"
+
+class TestVaspConverter:
+    "Test Vasp Converter"  
+
+    def test_singlepoint_conversion(self, single_point_vasp):
```
dormrod, line 29: These tests actually require AMS to run the rkf generation.

Import `skip_if_no_ams_installation` from `test_helpers` as in other files, and call it in these tests to skip them on GitHub.

unit_tests/test_converters.py
```diff
@@ -0,0 +1,53 @@
...
+        shutil.rmtree(rkf_dir)
+    
+    def test_NpT_conversion(self, npt_vasp):
+        rkf_dir = Path(vasp_output_to_ams(str(npt_vasp)))
+        reader = KFReader(rkf_dir/"ams.rkf")
+
+        assert reader.read("General", "task") == "moleculardynamics" 
+        assert reader.read("MDHistory", "Time(1)") == list(np.arange(0, 10, dtype=float))
+        
+        shutil.rmtree(rkf_dir)
```
dormrod, line 46: There could be an issue with this test in the following situation:

* test runs and fails the assert
* rkf is not cleaned up
* test is fixed
* test is re-run, default behaviour is not to overwrite the rkf so it still fails

So either, pass `overwrite = True` to the call, or also check if a file already exists at the start of the test and remove it

dormrod (changes requested):
A nice change overall 👏 Just some small things to tidy up

Fra-me:
Addressed all review comments + ran black

dormrod (approved):

PR #405, opened 2026-06-12
Rosa bulo/pdb and mol2

rebadam:
Small bugfix for Mol2 reading: Before, Plams would first write bond orders as floats in the mol2 file, but then when reading, floats were not accepted. I now changed the reading code to accept bond orders as floats in the mol2 format.

PDB file writing would not place any default values in the residue, resnum, occ, and fix columns if they were not specifically assigned as properties to the Molecule object. This resulted in a strangely empty format that would not be accepted by any software I know. I changed to print default values in those four columns.

### DIFF
--- a/src/scm/plams/mol/molecule.py
+++ b/src/scm/plams/mol/molecule.py
@@ -3178,6 +3178,8 @@ class Molecule:
 
             elif mode[0] == "BOND":
                 spl = line.split()
+                if "." in spl[3]:
+                    spl[3] = f"{int(float(spl[3]))}"
                 if len(spl) < 4:
                     raise FileError(f"readmol2: Error in {f.name} line {i+1}: not enough values in line")
                 try:
@@ -3274,8 +3276,13 @@ class Molecule:
         pdb = PDBHandler()
         for i, at in enumerate(self.atoms):
             pdbatom = PDBAtom()
+            pdbatom.name = "%-2s" % (at.symbol.upper())
             pdbatom.coords = at.coords
             pdbatom.element = at.symbol.upper()
+            pdbatom.res = "LIG"
+            pdbatom.resnum = "0"
+            pdbatom.occ = "1.00"
+            pdbatom.fix = "0.00"
             if "pdb" in at.properties:
                 if "res" in at.properties.pdb:
                     pdbatom.res = at.properties.pdb.res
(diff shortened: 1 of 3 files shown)


src/scm/plams/mol/molecule.py
```diff
@@ -3278,15 +3280,24 @@ def writepdb(self, f: IO, **other: Any) -> None:
         pdb = PDBHandler()
         for i, at in enumerate(self.atoms):
             pdbatom = PDBAtom()
+            pdbatom.name = "%-2s" % (at.symbol.upper())
```
dormrod, line 3284:
```suggestion
            pdbatom.name = f"{at.symbol.upper():<2}"
```

src/scm/plams/mol/molecule.py
```diff
@@ -3278,15 +3280,24 @@ def writepdb(self, f: IO, **other: Any) -> None:
...
                 if "res" in at.properties.pdb:
                     pdbatom.res = at.properties.pdb.res
                 if "resnum" in at.properties.pdb:
                     pdbatom.resnum = at.properties.pdb.resnum
                 if "name" in at.properties.pdb:
                     pdbatom.name = at.properties.pdb.name
+                if "occ" in at.properties.pdb:
+                    pdbatom.occ = at.properties.pdb.occ
+                if "fix" in at.properties.pdb:
+                    pdbatom.occ = at.properties.pdb.fix
```
dormrod:
```suggestion
                    pdbatom.fix = at.properties.pdb.fix
```

this looks like it should be `pdbatom.fix`?

src/scm/plams/mol/molecule.py
```diff
@@ -3182,6 +3182,8 @@ def readmol2(self, f: IO, **other: Any) -> None:
 
             elif mode[0] == "BOND":
                 spl = line.split()
+                if "." in spl[3]:
```
dormrod: This looks like it should be below the:
```
                if len(spl) < 4:
                    raise FileError(f"readmol2: Error in {f.name} line
```
check below? Otherwise this will error with and index error before that file error is reached

dormrod (changes requested):

dormrod:
Also please mention your fix in the changelog! https://github.com/SCM-NV/PLAMS/blob/trunk/CHANGELOG.md

src/scm/plams/mol/molecule.py
rebadam: Sure, done

src/scm/plams/mol/molecule.py
rebadam: Yes, you are correct, thanks!

src/scm/plams/mol/molecule.py
rebadam: Indeed!

dormrod (approved):

PR #407, opened 2026-06-15
plams.view xvfb: fix display_number 0 truthiness SO--

mhellstr:
This fix lets me use amsview_xvfb in a podman container. Also needed on fix2026.


dormrod (approved):
Good catch

PR #415, opened 2026-08-06
Add matplotlib energy landscape plot helper with selectable layouts Unit test included.

nfaguirrec:
Now we can do these things :)

plot_energy_landscape

Also, including the molecules:

plot_energy_landscape_with_molecules

For surfaces, it also works!

plot_energy_landscape_surfaces


unit_tests/test_tools_plot.py
```diff
@@ -564,3 +565,189 @@ def test_plot_work_function(run_calculations, rkf_tools_plot):
...
+
+# ----------------------------------------------------------
+# Testing plot_energy_landscape layouts for surfaces
+# ----------------------------------------------------------
+@image_comparison(
+    baseline_images=["plot_energy_landscape_surfaces_layouts"],
+    remove_text=True,
+    extensions=["png"],
+    style="mpl20",
+    tol=20,
```
dormrod, line 652: These can be a bit flakey, but tol of 20 is quite high so should be ok...

src/scm/plams/tools/plot.py
```diff
@@ -961,3 +967,417 @@ def plot_work_function(
         )
 
     return ax
+
+
+@requires_optional_package("matplotlib")
+def plot_energy_landscape(
```
dormrod, line 973: Re-export in top-level plams? 

https://github.com/SCM-NV/PLAMS/blob/trunk/src/scm/plams/__init__.py#L151

src/scm/plams/tools/plot.py
```diff
@@ -961,3 +967,417 @@ def plot_work_function(
...
+    ax: Optional["plt.Axes"] = None,
+    unit: str = "eV",
+    landscape_width: float = 0.45,
+    spacing: float = 1.0,
+    ts_color: str = "red",
+    min_color: str = "black",
+    connector_color: str = "black",
+    connector_linestyle: Any = (0, (4, 4)),
+    label_states: bool = True,
+    layout: str = "auto",
```
dormrod: Can we help out the user here: `Literal["auto", "dfs", "bfs", "longest_path", "force", "crossings"]`?

src/scm/plams/tools/plot.py
```diff
@@ -961,3 +967,417 @@ def plot_work_function(
...
+                molecule_width,
+                molecule_height,
+            )
+            inset_ax = ax.inset_axes(inset_bounds, transform=ax.transData)
+            inset_ax.set_facecolor("none")
+            inset_ax.patch.set_alpha(0.0)
+            state_label = getattr(state, "display_id", state.id)
+            state_plot_kwargs = dict(molecule_plot_kwargs)
+            state_plot_kwargs.update(molecule_plot_kwargs_by_state.get(state_label, {}))
+            plot_molecule(state.molecule, ax=inset_ax, keep_axis=False, **state_plot_kwargs)
```
dormrod: You might be able to use `view` method here instead, if you wanted AMSview style molecules instead of ASE style molecules

dormrod (commented):
Looks very nice! 

A tiny nit-pick is that it looks to me like the numbers cover some of the molecules with a white square, but this is very minor indeed ;)

Also remember to update the CHANGELOG!

unit_tests/test_tools_plot.py
nfaguirrec: I just copied it from the other tests. So far, it works. But it is a great point to keep in mind, in case at some point the tests fail

src/scm/plams/tools/plot.py
nfaguirrec: Well, every plot_* function in src/scm/plams/tools/plot.py has the same decorator, 10 in total. I thought it was on purpose. 

```
$ grep -H -A1 'requires_optional_package("matplotlib")' src/scm/plams/tools/plot.py 
src/scm/plams/tools/plot.py:@requires_optional_package("matplotlib")
src/scm/plams/tools/plot.py-def plot_band_structure(
--
src/scm/plams/tools/plot.py:@requires_optional_package("matplotlib")
src/scm/plams/tools/plot.py-def plot_phonons_band_structure(
--
src/scm/plams/tools/plot.py:@requires_optional_package("matplotlib")
src/scm/plams/tools/plot.py-def plot_phonons_dos(
--
src/scm/plams/tools/plot.py:@requires_optional_package("matplotlib")
src/scm/plams/tools/plot.py-def plot_phonons_thermodynamic_properties(
--
src/scm/plams/tools/plot.py:@requires_optional_package("matplotlib")
src/scm/plams/tools/plot.py-@requires_optional_package("ase")
--
src/scm/plams/tools/plot.py:@requires_optional_package("matplotlib")
src/scm/plams/tools/plot.py-def plot_image_grid(
--
src/scm/plams/tools/plot.py:@requires_optional_package("matplotlib")
src/scm/plams/tools/plot.py-def plot_correlation(
--
src/scm/plams/tools/plot.py:@requires_optional_package("matplotlib")
src/scm/plams/tools/plot.py-def plot_msd(
--
src/scm/plams/tools/plot.py:@requires_optional_package("matplotlib")
src/scm/plams/tools/plot.py-def plot_work_function(
--
src/scm/plams/tools/plot.py:@requires_optional_package("matplotlib")
src/scm/plams/tools/plot.py-def plot_energy_landscape(
```

src/scm/plams/tools/plot.py
nfaguirrec: I have added the parameter ``molecule_plot_backend: Literal["view", "plot_molecule"] = "view"`` to the ``plot_energy_landscape`` function. Now, the user can choose which backend they prefer

nfaguirrec: By the way, I just realized that ``molecule_plot_backend="view"``cannot be tested in CI because ams is not installed there

nfaguirrec: I also added an example ``examples/EnergyLandscape/EnergyLandscape.ipynb`` where I show how to use these new functions, including the use of `view` and `plot_molecule`

nfaguirrec: I also added this option (highlighting states):

image

src/scm/plams/tools/plot.py
nfaguirrec: I added this in the function's documentation:

- Use ``"auto"`` to compare the available strategies and pick the one with the cleanest connector pattern.
- Use ``"dfs"`` to follow one branch deeply before backtracking, which can resemble a reaction-path view.
- Use ``"bfs"`` to expand level by level from one endpoint, keeping nearby states grouped together.
- Use ``"longest_path"`` to place the main backbone of the network first and then attach side branches around it.
-  Use ``"force"`` to apply a simple force-based relaxation that spreads states while reducing visual crowding.
- Use ``"crossings"`` to minimize connector crossings directly, which can help for dense networks.

nfaguirrec:
> Looks very nice!
> 
> A tiny nit-pick is that it looks to me like the numbers cover some of the molecules with a white square, but this is very minor indeed ;)
> 
> Also remember to update the CHANGELOG!

Thanks!
In those cases, you can use `molecule_y_offset` to hide the white squares :)
I already updated the CHANGELOG!

src/scm/plams/tools/plot.py
dormrod: I meant add to the top level __init__.py (see link above) so it can be imported directly from plams

src/scm/plams/tools/plot.py
dormrod: I would use the `Literal` type hint as these are the only options, so users with IDEs will get highlighting for an incorrect option

dormrod (approved):

src/scm/plams/tools/plot.py
nfaguirrec: Good Idea! Done! :)

PR #417, opened 2026-08-13
add plot_molecule_counts and AMSResults.get_molecule_count_history() …

mhellstr:
Create a reasonable interface to plot those quantities from ams.rkf files (similar to AMSmovie).

I'll also update https://www.scm.com/doc.trunk/PythonExamples/bond-counts-rkf/  to use the new interface.


dormrod (commented):
You might want to add to the top level __init__.py (https://github.com/SCM-NV/PLAMS/blob/trunk/src/scm/plams/__init__.py#L151)

PR #421, opened 2026-09-03
Fix PLUMED Input block serialization SO--

anastasia-krch:
Made PLUMED input block work with active learning


dormrod (approved):
Looks good, thanks!

PR #426, opened 2026-09-22
Wei lin chen/crs builder review cleanup

WLChen1118:
For instructions on using the property-aware CRS input builder, see the
https://github.com/SCM-NV/PLAMS/blob/WeiLinChen/crs-builder-review-cleanup/doc/source/interfaces/crs.rst

I’m not sure whether class CRSInputBuilder.to_inputs() is the structure we want, but it provides a convenient way to convert the builder state into a typed scm.inputs.CRS model. This also allows users to configure additional CRS options that are not exposed by the property-specific builder.

### DIFF
--- a/src/scm/plams/interfaces/adfsuite/crs.py
+++ b/src/scm/plams/interfaces/adfsuite/crs.py
@@ -139,6 +139,13 @@ class CRSResults(SCMResults):
         except:
             nstruct = ncomp
 
+        try:
+            nmoment = cast(int, self.readkf(section, "nmoment"))
+            nhb_moment = cast(int, self.readkf(section, "nhb_moment"))
+        except:
+            nmoment = 0
+            nhb_moment = 0
+
         np_dict: Dict[str, Any] = {"section": section}
         np_dict["ncomp"] = ncomp
         chunk_length = 160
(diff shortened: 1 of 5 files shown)


src/scm/plams/interfaces/adfsuite/crs.py
```diff
@@ -28,12 +48,31 @@ def section(self) -> str:
...
+            header = property_settings.get("_h") if property_settings is not None else input_data.get("t")
+
+        if not isinstance(header, str) or not header:
+            raise ValueError("Could not determine the CRS property section from the job input")
 
-            return self._section
+        self._section = header.upper()
+        return self._section
+
+        # except AttributeError:
```
dormrod: (remove dead code)

src/scm/plams/interfaces/adfsuite/crs.py
```diff
@@ -28,12 +48,31 @@ def section(self) -> str:
...
         except AttributeError:
-            try:
-                self._section = self.job.settings.input.property._h.upper()
-            except AttributeError:
-                self._section = self.job.settings.input.t.upper()
+            pass
+
+        input_data = self.job.settings.input
+
+        from scm.inputs import CRS
```
dormrod, line 56: `@requires_optional_package("scm.inputs")` technically optional even through present in amspython

robertrueger (approved):
The user interface looks pretty nice. I think the implementation would have been much more straightforward if we had the new new builder classes based on pydantic, as that would do all the validation based on the type-hints automatically. But that alone is probably not a good reason to pull in pydantic as a new dependency of PLAMS ...

WLChen1118:
@dormrod  In <[34a03ba](https://github.com/SCM-NV/PLAMS/pull/426/commits/34a03ba5bc9502fc7124f2df3483d3c719f63c51)>, I added a minimal JSON fixture so the CRS input builder tests can run without an AMS installation. Does this approach look okay to you?

